PR #4, opened 2015-02-19
More server tweaks

mattheath:
- [x] Tweak example handler name to drop the 'handler': `func Hello(req server.Request) (server.Response, error)`
 - [x] `s/server.Initialise/server.Init` because s/z :wink: 
 - [x] `s/Rabbit/AMQP` in a few more places
 - [x] Publish an error back to the client on error, rather than not responding
 - [x] Updated imports in all packages to the new location.
 - [x] Change handler interface to return a `server.Response` interface which must implement an `Encode()` method

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -6,11 +6,11 @@ import (
 	"os"
 	"time"
 
+	"github.com/b2aio/typhon/rabbit"
 	log "github.com/cihub/seelog"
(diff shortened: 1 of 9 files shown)


jonashuckestein:
LGTM!

vinceprignano:
Rocks :rocket:

PR #6, opened 2015-02-23
re-implement stub server

jonashuckestein:


server/request.go
```diff
@@ -25,6 +29,26 @@ func NewAMQPRequest(delivery *amqp.Delivery) *AMQPRequest {
 
 // RabbitMQ / AMQP fields
 
+func (r *AMQPRequest) ServiceName() string {
+	routingKey := r.RoutingKey()
+	lastDotIndex := strings.LastIndex(routingKey, ".")
+	if lastDotIndex == -1 {
+		return routingKey
+	} else {
```
mattheath: You can get rid of the else here, and just return after

server/request.go
```diff
@@ -25,6 +29,26 @@ func NewAMQPRequest(delivery *amqp.Delivery) *AMQPRequest {
...
+		return routingKey[:lastDotIndex]
+	}
+}
+
+func (r *AMQPRequest) Endpoint() string {
+	routingKey := r.RoutingKey()
+	lastDotIndex := strings.LastIndex(routingKey, ".")
+	if lastDotIndex == -1 {
+		return ""
+	} else {
```
mattheath: Same here - can remove else as you're returning early

mattheath:
Looks great, couple of minor tweaks and we're good to merge. :+1:

PR #7, opened 2015-02-24
Platform errors

mattheath:
Add some generic error types to be passed between services. Rudimentary for now, but allows us to update the handler function signature to the below, and pass back encapsulated errors from client and server.

```go
HandleRequest(req Request) (Response, errors.Error)
```


errors/errors.go
```diff
@@ -0,0 +1,130 @@
...
+	Type() ErrorType
+}
+
+// ErrorType is an enumerated type of error
+type ErrorType string
+
+const (
+	// ErrUnknown indicates an unknown type of error
+	// @todo should this just be mapped to an internal server error?
+	ErrUnknown = ErrorType("UnknownError")
```
jonashuckestein: I agree with the TODO. As far as Im concerned everything unexpected is an InternalServerError

PR #8, opened 2015-02-24
Add RequestType and ResponseType to Endpoints and refactor test helpers ...

jonashuckestein:
...to not use testify.TestSuite


server/request.go
```diff
@@ -10,14 +10,21 @@ import (
 type Request interface {
 	context.Context
 
-	Body() []byte
+	// Payload are the actual bytes returned from the transport
+	Payload() []byte
+
+	// Body is the Unmarshalled `Payload()`. If `RequestType()` is set on
+	// the `Endpoint`, we can attempt to unmarshal it for you
```
mattheath: Presumably this would mean an end user should use `Payload()` rather than `Body()` if they haven't set the `RequestType`?

test/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+	require.NoError(t, err)
+	resp, err := endpoint.HandleRequest(server.NewAMQPRequest(&amqp.Delivery{
+		Body: reqBytes,
+	}))
+	if err != nil {
+		return err
+	}
+	respBytes, err := resp.Encode()
+	require.NoError(t, err)
+	err = proto.Unmarshal(respBytes, respProto)
```
mattheath: This encode/decode step is used because we get a `server.Response` back from the handler?

test/stub_server.go
```diff
@@ -54,24 +57,31 @@ func NewStubServer(testSuite *RabbitTestSuite) *StubServer {
...
+}
 
-	// determine which endpoint to use
-	for _, stub := range stubServer.stubs {
-		if stub.ServiceName == req.ServiceName() && stub.Endpoint == req.Endpoint() {
-			return stub.Handler(req)
-		}
-	}
+// StubError is a convenience method to stub out a service error
+func (stubServer *StubServer) StubError(serviceName, endpoint string, err error) {
```
mattheath: Convention for `func` signatures here would probably be the below, but not important :smile: 

```go
func (s *StubServer) StubError(serviceName, endpoint string, err error) {
```

mattheath:
:ship:

PR #10, opened 2015-02-25
Client Refactoring

mattheath:
~~This is still very much a work in progress~~

 - [x] Make client lazily connect when used for the first time
 - [x] Change `Dispatcher` which tracked inflight requests to a more appropriately named `inflightRegistry`
 - [x] Change client interface to take a `service` and `endpoint` rather than a `routingKey` which was tightly coupled to the rabbitmq implementation
 - [x] Make `client.Request` take a `context.Context` as its first param
 - [x] Add `ScopedRequest` to a `server.Request` which makes a client request within the context of the server request - this should be used in most cases
 - [ ] ~~Implement underlying usage of context to pass information to the receiving service~~
 - [ ] ~~Tidy up client so it better initialises itself, and reconnects if disconnected~~


mattheath:
@jonashuckestein good to go. the other changes will be behind the interface, so can be implemented later. :ship: :ship: :shipit:

client/rabbitmq.go
```diff
@@ -0,0 +1,140 @@
...
+		connection: rabbit.NewRabbitConnection(),
+		replyTo:    fmt.Sprintf("replyTo-%s", uuidQueue.String()),
+	}
+}
+
+func (c *RabbitClient) Init() {
+	select {
+	case <-c.connection.Init():
+		log.Info("[Client] Connected to RabbitMQ")
+	case <-time.After(10 * time.Second):
```
jonashuckestein: For the next version: let's make this significantly less and/or configurable . In development services that are going to die anyway should die quickly

client/rabbitmq.go
```diff
@@ -0,0 +1,140 @@
...
+		log.Critical("[Client] Failed to connect to RabbitMQ")
+		os.Exit(1)
+	}
+	c.initConsume()
+}
+
+func (c *RabbitClient) initConsume() {
```
jonashuckestein: Comment

client/rabbitmq.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	go func() {
+		log.Infof("[Client] Listening for deliveries on %s", c.replyTo)
+		for delivery := range deliveries {
+			go c.handleDelivery(delivery)
+		}
+	}()
+}
+
+func (c *RabbitClient) handleDelivery(delivery amqp.Delivery) {
```
jonashuckestein: Comment

client/rabbitmq.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+	select {
+	case channel <- delivery:
+	default:
+		log.Errorf("[Client] Error in delivery for correlation %s", delivery.CorrelationId)
+	}
+}
+
+func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, req proto.Message, res proto.Message) error {
```
jonashuckestein: Comment

jonashuckestein:
This looks good. We'll fix the comments the next go around

client/rabbitmq.go
mattheath: :+1:

mattheath: This is just used while waiting for a connection - the library underneath should be continually trying to reconnect during this.

PR #12, opened 2015-03-04
Separate test server, allow remote Close()

mattheath:
Allow `Close()` to be called on the server, terminating the AMQP connection, and shutting down the server listening loop.

Also move the test server to the test package, so we can reuse it from other packages

### DIFF
--- a/test/server.go
+++ b/test/server.go
@@ -1,0 +1,27 @@
+package test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/b2aio/typhon/server"
+)
+
+// InitServer for testing
+func InitServer(t *testing.T, name string) server.Server {
+	// Initialize our Server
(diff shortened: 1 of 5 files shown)


test/server.go
```diff
@@ -0,0 +1,27 @@
...
+		Name:        name,
+		Description: "Example service",
+	})
+
+	go server.Run()
+
+	select {
+	case <-server.NotifyConnected():
+	case <-time.After(1 * time.Second):
+		t.Fatalf("StubServer couldn't connect to RabbitMQ")
```
jonashuckestein: This isn't a StubServer I think

jonashuckestein:
Yes please! Very important fix :)

test/server.go
mattheath: Correct!

PR #13, opened 2015-03-04
Error Propagation

mattheath:
Follows on from #12 which pulls out the test server allowing easy reuse.

Adds error propagation using protobufs between services, marshaling on the server end, and unmarshaling back to the original error type on the client service.

Uses a `Content-Encoding` header on rabbitMQ to communicate which protobuf to use for deserialisation.


jonashuckestein:
:+1:

jonashuckestein:
ci rebuild

PR #14, opened 2015-03-08
Fix typo and potential panics

mattheath:
Writing to a closed channel results in a panic. This means if a client times out, and closes its reply channel, but a reply _was_ sent, receiving this causes a runtime panic.

By not closing the channel we allow a response to come back, and be written into the channel, at which point it will fall out of scope at both ends and the garbage collector will pick it up at some point.

Separately we should be manually collecting timed out responses from the inflight registry after a period of time - otherwise in the case we never receive a response the channel will never be garbage collected.


jonashuckestein:
Looks good. Once typhon matures a little bit I'd like to get into the habit of writing regression tests for this kind of thing

PR #15, opened 2015-03-11
add public and private context to errors, client code registry

jonashuckestein:
Let's add actual client codes to the registry in coordination with james tomorrow or on friday


errors/errors.go
```diff
@@ -51,6 +58,40 @@ func (p *ServiceError) Description() string {
 	return ""
 }
 
+// ClientCode returns the error code that the client uses to display.
+// There are much fewer client codes than actual error codes. The mapping
+// from code to client code is done in `errorFactory` using data from `client_codes.go`
+func (p *ServiceError) ClientCode() int {
+	if p != nil {
+		return p.clientCode
```
mattheath: Think you also want to check `p.clientCode != 0` and return the default if its not initialised.

PR #17, opened 2015-03-17
Some cleanup

jonashuckestein:
- new errors.Error type with Code (an integer), Message, Stack, PublicContext and PrivateContext
 - change endpoint handler signature to return proto.Message instead of server.Response interface
 - log stacktraces for errors


errors/errors.go
```diff
@@ -1,179 +1,141 @@
...
+func Unauthorized(message string, context ...map[string]string) *Error {
+	return errorFactory(ErrUnauthorized, message, context...)
 }
 
-// errorConstructor returns a `ServiceError` of the specified type. The main work
+// errorConstructor returns a `ServiceError` of the specified code. The main work
 // consists of managing the map[string]string at the end of the arguments list.
 // In practice we only ever pass in two: the first one is public and will be sent to the client,
 // the second one is private and can contain internal information that is useful for debugging
 // @todo get a stack trace and stick it in the private context
```
mattheath: this todo is presumably now done?

server/endpoint.go
```diff
@@ -81,3 +57,21 @@ func cloneTypedPtr(reqType interface{}) interface{} {
 	reflectValue := reflect.New(reflect.TypeOf(reqType).Elem())
 	return reflectValue.Interface()
 }
+
+// enrichError converts an error interface into errors.ServiceError and attaches
```
mattheath: `s/errors.ServiceError/errors.Error`

test/error_passing_test.go
```diff
@@ -54,20 +51,18 @@ func TestErrorPropagation(t *testing.T) {
...
-	// Type assert this to a service error
-	svcErr, ok := err.(*errors.ServiceError)
-	require.Equal(t, true, ok)
-
 	// Assert our error matches
 	require.NotNil(t, err)
-	assert.Equal(t, errorCode, svcErr.Code())
-	assert.Equal(t, errorDescription, svcErr.Description())
-	assert.Equal(t, errorDescription, svcErr.Error())
+	typhonErr := err.(*errors.Error)
```
mattheath: we should probably catch a type assertion failure here, and _require_ it was `ok`

mattheath:
Really like this, couple of very minor things, then good to :ship:

mattheath:
:+1:

PR #18, opened 2015-03-19
Fix various bugs including 100% cpu usage on connection close

mattheath:
- [x] Fix infinite blank delivery bug when the AMQP connection closed - caused 100% cpu usage
 - [x] Add _much_ better logging. Some parts are verbose, but set at `trace` and `debug` levels
 - [x] Fix message delivery acknowledgement - we weren't ack'ing messages, we now auto ack
 - [x] Catch errors on connection, channel & queue declaration etc, and return them rather than swallowing them
 - [x] Fail immediately in the client if not connected


jonashuckestein:
This is great, Thanks!

PR #20, opened 2015-03-23
Refactoring to support RPC

mattheath:
Refactoring to make the interfaces a little more generic, and support raw requests in multiple encodings - allowing us to support JSON on the wire between services, and therefore push raw JSON requests in from an API.

## Client
 - [x] Add `client.Request` and `client.Response` interfaces, with concrete implementations
 - [x] Change `Call` func to `Req` (can't be called `Request` as it conflicts with the interface above)
 - [x] Add `CustomReq` which takes a raw `Request` and returns a `Response`
 - [x] Refactor underlying client request code into a generalised `do` method. This is less specific to AMQP, but still has some mentions of it
 - [x] Pass additional headers through rabbitmq to the server, including the content type (proto/json) of the message
 - [x] ~~Push AMQP client details down into the `rabbit` package~~
 - [x] ~~Move response type down into `rabbit` pkg, leaving interface in `client` pkg, and unmarshal there~~

## Server
 - [x] Unmarshal additional headers from client
 - [x] Use `Content-Type` header to determine unmarshaling func, now supports both proto and json
 - [x] Simplify service and endpoint identification by using delivery headers
 - [x] Marshal response back to client in the content-type they sent
 - [x] ~~Tidy up responding with a response interface / type~~

### DIFF
--- a/client/rabbitmq.go
+++ b/client/rabbitmq.go
@@ -87,7 +87,7 @@ func (c *RabbitClient) handleDelivery(delivery amqp.Delivery) {
 	}
 }
 
-func (c *RabbitClient) Request(ctx context.Context, serviceName, endpoint string, req proto.Message, resp proto.Message) error {
+func (c *RabbitClient) Req(ctx context.Context, serviceName, endpoint string, req proto.Message, resp proto.Message) error {
 
 	// Ensure we're initialised, but only do this once
 	//
--- a/client/rabbitmq.go
+++ b/client/rabbitmq.go
@@ -87,7 +87,7 @@ func (c *RabbitClient) handleDelivery(delivery amqp.Delivery) {
 	}
 }
 
-func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, req proto.Message, resp proto.Message) error {
+func (c *RabbitClient) Request(ctx context.Context, serviceName, endpoint string, req proto.Message, resp proto.Message) error {
 
 	// Ensure we're initialised, but only do this once
 	//
--- a/client/rabbitmq.go
+++ b/client/rabbitmq.go
@@ -155,9 +155,15 @@ func (c *RabbitClient) do(req Request) (Response, error) {
 		},
 	}
 
+	// Attempt to publish through our connection
+	// @todo refactor this to not know about rabbitmq internals
 	err := c.connection.Publish(rabbit.Exchange, routingKey, message)
 	if err != nil {
 		log.Errorf("[Client] Failed to publish %s to '%s': %v", req.Id(), routingKey, err)
+
+		// Remove from inflight requests
+		_ = c.inflight.pop(req.Id())
+
 		return nil, errors.Wrap(err) // @todo custom error code
 	}
 
--- a/client/rabbitmq.go
+++ b/client/rabbitmq.go
@@ -183,7 +183,7 @@ func deliveryIsError(delivery amqp.Delivery) bool {
 		return true
 	}
 
-	if encoding == "" || encoding == "ERROR" {
+	if encoding == "" || encoding == "error" {
 		return true
 	}
 
--- a/client/rabbitmq.go
+++ b/client/rabbitmq.go
@@ -156,10 +156,13 @@ func (c *RabbitClient) buildRoutingKey(serviceName, endpoint string) string {
 // handleResponse returned from a service by marshaling into the response type,
 // or converting an error from the remote service
 func handleResponse(delivery amqp.Delivery, resp proto.Message) error {
+
+	serverResp := deliveryToResponse(delivery)
+
 	// deal with error responses, by converting back from wire format
-	if deliveryIsError(delivery) {
+	if serverResp.IsError() {
 		p := &pe.Error{}
-		if err := proto.Unmarshal(delivery.Body, p); err != nil {
+		if err := proto.Unmarshal(serverResp.Payload(), p); err != nil {
 			return errors.BadResponse(err.Error())
 		}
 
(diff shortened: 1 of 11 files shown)


mattheath:
Closed off the last couple of items - this PR is big enough already. Also want to tidy up the connection logic - at the moment the client doesn't reconnect on failure, or disconnection, whereas we should reattempt on requests. I'll do this separately at some point.

client/rabbitmq.go
```diff
@@ -87,7 +88,42 @@ func (c *RabbitClient) handleDelivery(delivery amqp.Delivery) {
...
 }
 
-func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, req proto.Message, resp proto.Message) error {
+func (c *RabbitClient) Req(ctx context.Context, service, endpoint string, req proto.Message, resp proto.Message) error {
+
+	// Build request
+	payload, err := proto.Marshal(req)
+	if err != nil {
+		log.Errorf("[Client] Failed to marshal request: %v", err)
+		return errors.Wrap(err) // @todo custom error code
```
jonashuckestein: Now that we've refactored this the way we like it, we could give all errors coming out of typhon a unique error code. We could say any error code < 1000 is from typhon

client/rabbitmq.go
```diff
@@ -97,91 +133,90 @@ func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, r
...
+		service:         service,
+		endpoint:        endpoint,
+		payload:         delivery.Body,
 	}
+}
 
-	if encoding == "" || encoding == "ERROR" {
-		return true
+// unmarshalErrorResponse from our wire format to a typhon error
+func unmarshalErrorResponse(resp Response) *errors.Error {
```
jonashuckestein: Unfortunately, even in this special case, we should return an `error` interface. Otherwise the line `errors.Unmarshal(p)` may return a "corrupted" `nil` that we can never return from a method as an `error` interface

client/rabbitmq.go
```diff
@@ -97,91 +133,90 @@ func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, r
...
 	message := amqp.Publishing{
-		CorrelationId: correlation.String(),
+		CorrelationId: req.Id(),
 		Timestamp:     time.Now().UTC(),
-		Body:          requestBody,
+		Body:          req.Payload(),
 		ReplyTo:       c.replyTo,
+		Headers: amqp.Table{
+			"Content-Type":     req.ContentType(),
+			"Content-Encoding": "request",
```
jonashuckestein: What will this be used for? Is it mostly for debugging?

client/rabbitmq.go
```diff
@@ -97,91 +133,90 @@ func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, r
...
 	case delivery := <-replyChannel:
-		log.Debugf("[Client] Response received for %s from %s", correlation.String(), routingKey)
-		return handleResponse(delivery, resp)
+		log.Debugf("[Client] Response received for %s from %s", req.Id(), routingKey)
+		rsp := deliveryToResponse(delivery)
+		if rsp.IsError() {
+			return nil, unmarshalErrorResponse(rsp)
+		}
+		return rsp, nil
 	case <-time.After(defaultTimeout):
```
jonashuckestein: We should move this into `Request` as soon as possible and put it into the request headers, too. These cascading timeouts are getting a bit tedious to debug, especially now that I get otherwise useful errors from the API. How are we going to make ScopedRequests with a timeout? With a second helper method called `ScopedRequestWithTimeout`? Are we going to automatically set the timeout of scoped requests to be the parent's timeout less some fudge latency amount?

client/request.go
```diff
@@ -0,0 +1,93 @@
+package client
+
+import (
+	log "github.com/cihub/seelog"
+	"github.com/nu7hatch/gouuid"
+
+	"github.com/b2aio/typhon/errors"
+)
+
+type Request interface {
```
jonashuckestein: I still don't like interfaces with just one implementation that contains mostly trivial getter methods. I get that it protects developers from messing with data, but then again we're all grown ups and don't like boilerplate.

FWIW most types I remember seeing in the go standard library don't bother to hide their internal fields behind a read-only interface (for example http.Request)

client/rabbitmq.go
```diff
@@ -97,91 +133,90 @@ func (c *RabbitClient) Call(ctx context.Context, serviceName, endpoint string, r
...
+		},
 	}
 
-	err = c.connection.Publish(rabbit.Exchange, routingKey, message)
+	err := c.connection.Publish(rabbit.Exchange, routingKey, message)
 	if err != nil {
-		log.Errorf("[Client] Failed to publish %s to '%s': %v", correlation.String(), routingKey, err)
-		return errors.Wrap(err) // @todo custom error code
+		log.Errorf("[Client] Failed to publish %s to '%s': %v", req.Id(), routingKey, err)
+		return nil, errors.Wrap(err) // @todo custom error code
```
jonashuckestein: Super minor issue, but I think this return may cause a tiny leak in the inflight registry

client/rabbitmq.go
jonashuckestein: Never mind, found it. Used to figure out if something is an error. I'd perhaps add a comment

server/rabbitmq.go
```diff
@@ -140,20 +142,30 @@ func (s *AMQPServer) handleRequest(delivery amqp.Delivery) {
 		return
 	}
 
-	body, err := proto.Marshal(resp)
+	// Marshal the response
+	// @todo we're currently always marshaling errors as proto
```
jonashuckestein: In the short term we can address that on the API level. There already is a method to take an error and render it in json

client/rabbitmq.go
mattheath: I'd like these to be automatically discovered through some service discovery method ideally, but capped at the parents timeout minus approximate network latency (which we could measure!).

client/rabbitmq.go
mattheath: Yes all the responses are sent back with either `response` or `error`, thought it made sense to standardise this with a `request` type. Also if we add heartbeart/pings these could use this field to identify themselves.

jonashuckestein:
Awesome PR! It's a tad complex for my taste, but who knows, perhaps we do end up with many transport/encoding combinations after all ;)

This is excellent groundwork for some typhon next steps I can think of:

 - [ ] separate out `server.Request` and `ctx`
 - [ ] generate a trace ID and pass the parent request in using rabbit headers
 - [ ] implement timeouts properly
 - [ ] add error codes throughout

client/request.go
mattheath: This is mainly because I'm planning on removing the `response` concrete type from the `client` package, and pushing it into the underlying transport where we unmarshal from the delivery into a `Response` - allowing us to return a `Response` interface. Having only one of `Request` and `Response` as an interface made the `Req` method interfaces suck.

client/rabbitmq.go
mattheath: :+1: Will ticket up.

client/rabbitmq.go
mattheath: Good catch! :flushed:

PR #21, opened 2015-03-25
Handle unmarshaling nil requests

mattheath:
Handle unmarshaling nil requests, by just using a blank request. This means if you pass a blank string `""` rather than `"{}"` with a json content type the behaviour will now be the same.


jonashuckestein:
Good good

PR #22, opened 2015-03-31
increase defaultTimeout to 5s

tomblomfield:


mattheath:
Yep, let's do this until we make it configurable.

PR #24, opened 2015-04-14
Fix race condition notifying observers before consuming

mattheath:
Fix a race condition where observers were notified of the server connecting before the server started consuming messages (and both declared and bound to the exchange) thereby dropping messages sent immediately after. Also made this non blocking to prevent deadlocks if the observer had timed out or stopped listening.


jonashuckestein:
Nice find! :exclamation:

PR #26, opened 2015-04-14
Update inter-service routing

mattheath:
Change routing & binding on rabbitmq to send message just to the service's name, and bind with just the service name, rather than also specifying the endpoint in the routing key and binding on a wildcard. 

Bind on `service.thing` rather than `service.thing.#`
Send to `service.thing` rather than `service.thing.someendpoint`

This makes routing easier to reason about, and simpler, as we don't bind per endpoint anyway; and the `server` package deals with invalid endpoints anway.


tomblomfield:
lgtm

PR #27, opened 2015-04-14
Move service name to the stub server, rather than endpoints

mattheath:
Currently the stub server binds on a wildcard routing key, which means this can conflict with other 'real' services bound on the same exchange (useful in more complex tests). This changes the stub server to only support a single service per instance, though multiple instances can be created, with the service name provided on creation. Multiple endpoints for that service can then be registered & stubbed.

### DIFF
--- a/test/helpers.go
+++ b/test/helpers.go
@@ -12,9 +12,9 @@ import (
 
 var stubServer *StubServer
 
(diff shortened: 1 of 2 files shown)


jonashuckestein:
LGTM. Will merge this and https://github.com/b2aio/wearedev/pull/101 right after

PR #28, opened 2015-04-14
Add deadletter exchange

mattheath:
This sets up a dead-letter exchange, on connection as an _alternate exchange_ for the primary topic exchange. This will collect non-routable messages, which can then be inspected.

More information on alternate exchanges here: https://www.rabbitmq.com/ae.html

 * Given a named topic exchange of `b2a`, a deadletter exchange called `b2a.deadletter` will be created with a type of `fanout` (this ensures all bound queues get a copy of each message).
 * A durable queue called `deadletter` will be created and bound to this exchange
 * The `b2a.deadletter` exchange will be set as an _alternate exchange_ for the `b2a` exchange

This will cause all non routable messages on the `b2a` exchange to be routed onto the `b2a.deadletter` exchange, where they will drop into the durable `deadletter` queue.

_Note: As the deadletter queue is durable, it will need to be periodically emptied if a large number of messages are sent!_

### DIFF
--- a/rabbit/channel.go
+++ b/rabbit/channel.go
@@ -53,6 +53,15 @@ func (r *RabbitChannel) Publish(exchange, routingKey string, message amqp.Publis
 }
 
 func (r *RabbitChannel) DeclareExchange(exchange string) error {
+
+	args := amqp.Table{}
+
+	// If our deadletter exchange is enabled, set this as the alternate exchange
+	// See https://www.rabbitmq.com/ae.html for more info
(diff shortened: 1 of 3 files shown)


mattheath:
ci rebuild

mattheath:
ci rebuild

obeattie:
We should totally have a deadletter exchange at some point, but this is obviously incompatible with the refactor. Also, I think after discussions with @mattheath that it would be better to do this at the infrastructure automation level.

PR #30, opened 2015-04-21
initialise env vars inside new connection

mattheath:
Rather than in `init` so that we only do this when creating a connection, keeping log output until we try to connect.


jonashuckestein:
LGTM

PR #32, opened 2015-04-27
Fix panic

jonashuckestein:
The root cause of this panic is that typhon errors that come from from rabbitmq as protobufs can have a `nil` Private/PublicContext if they were marshaled with an empty map `map[string]string{}`

This solution breaks the symmetry of marshaling and unmarshaling to proto. It sucks, but unfortunately that's how protobuf works. There's no way to distinguish between something that was marshaled with an empty map and something that was marshaled with a nil map


mattheath:
:+1:

PR #33, opened 2015-04-29
First steps towards authentication

mattheath:
This is a first step, and could do with a lot of improvement, but allows us to register authorisation functions on each endpoint, and run these to grant access to the endpoints before execution.

 - [x] Pass a `Session` through client and servers with the request if set
 - [x] Add interface for `Credentials` which is similar to a basic set of oauth credentials
 - [x] Add `AuthenticationProvider` interface, and let this be registered with the server
 - [x] Add reference to the endpoint when registering with the server, so we can get the authentication provider and other future things from the server instance. _This is a bit crufty_.
 - [x] Add `Authorizer` func interface, ~~and add this as an optional field to an `Endpoint`~~
 - [x] Add a `DefaultAuthorizer` which will be used if no `Authorizer` is set on an endpoint (note: this can be overridden)
 - [x] If an `AuthenticationProvider` is available, retrieve the credentials, and pass these into the `Authorizer` to authorize access to the endpoint _before_ the handler is executed.
 - [x] Add basic interface for `User` and `Client` to the `Session` as a precursor to full oauth 2.0
 - [x] Make `AuthenticationProvider` available to the client, and correctly marshal session onto the wire
 - [x] Basic Tests

Improvements:
 * ~~I'm erring towards removing the session param from the `Authorizer` interface, and just passing through the context, then users can make helper funcs to extract the relevant info from the context.~~
 * ~~There is no high level concept of a user - just an access token. This isn't great.~~
 * No caching yet, ~~or passing through the recovered credentials, so each service would need to re-recover the authorisation details.~~


auth/auth.go
```diff
@@ -0,0 +1,26 @@
...
+type Credentials interface {
+	AccessToken() string
+	RefreshToken() string
+	Expiry() time.Time
+	Scopes() []string
+}
+
+// Authorizer provides an interface to validate authorization credentials
+// for access to resources, eg. oauth scopes, or other access control
+type Authorizer func(ctx context.Context, creds Credentials) error
```
jonashuckestein: nice use of the american z!

client/request.go
```diff
@@ -23,6 +23,9 @@ type Request interface {
 	// Request timeout
 	Timeout() time.Duration
 	SetTimeout(time.Duration)
+	// AccessToken for this request, used for authentication
+	AccessToken() string
+	SetAccessToken(string)
```
jonashuckestein: Is this idiomatic? I just implemented `SetTimeout` in an emergency with the minimum effort possible. 

Once we have more other things to set (parent request information, trace id, etc) constructing a `request` will become a long series of `setX()` calls. Might be worth making a richer constructor (or exporting the fields because this isn't Java...)

server/rabbitmq.go
```diff
@@ -59,6 +63,7 @@ func (s *AMQPServer) NotifyConnected() chan bool {
 }
 
 func (s *AMQPServer) RegisterEndpoint(endpoint *Endpoint) {
+	endpoint.server = s
```
jonashuckestein: If `server` is not set, the endpoint is invalid and might cause a nil pointer dereference in authenticateEndpointAccess. 

As endpoints get increasingly complex, I'd consider adding an endpoint validation step to the `Register` method and potentially returning an error here or on server.Run

jonashuckestein:
This looks good. Could you please add a basic test? It can be just another simple method in typhon_test.go

server/amqp_request.go
```diff
@@ -88,6 +92,10 @@ func (r *AMQPRequest) Interface() interface{} {
 	return r.delivery
 }
 
+func (r *AMQPRequest) AccessToken() string {
+	return r.accessToken
+}
+
 // Client implementation
 
 // ScopedRequest calls a service within the scope of the current request
```
jonashuckestein: I think it may be worth passing the access token to the child request even in this PR. Otherwise we can only play around with authentication on the "edge" services that get hit by the API

auth/auth.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"time"
+
+	"golang.org/x/net/context"
+)
+
+// AuthenticationProvider provides helper methods to convert tokens to sessions
+// using our own internal authorization services
+type AuthenticationProvider interface {
```
jonashuckestein: Are we expecting this interface to grow in size? If it'll always just be a RecoverSession method we could just pass around the method directly

jonashuckestein: The answer is yes we are ;)

server/rabbitmq.go
mattheath: Agreed. Arguably this should even `panic` as it would be executed on startup, and we want the error to not be ignored.

client/rabbitmq.go
```diff
@@ -155,6 +156,50 @@ func (c *RabbitClient) do(req Request) (Response, error) {
...
+	}
+	if accessToken != "" {
+		message.Headers["Access-Token"] = accessToken
+	}
+
+	// If we have a recovered session for the parent request, set it on the child request
+	// so that the called service doesn't have to call RecoverSession() again
+	if parentRequest != nil && parentRequest.HasRecoveredSession() &&
+		parentRequest.Server().AuthenticationProvider() != nil {
+		log.Debugf("Hullo %+v %+v %+v", parentRequest.HasRecoveredSession())
```
mattheath: Stray debug line

client/rabbitmq.go
```diff
@@ -155,6 +156,50 @@ func (c *RabbitClient) do(req Request) (Response, error) {
...
+		session, err := parentRequest.Session()
+		if err != nil {
+			return nil, err
+		}
+		sessionBytes, err := parentRequest.Server().AuthenticationProvider().MarshalSession(session)
+		if err != nil {
+			log.Warnf("Failed to marshal recovered session") // @todo log somewhere
```
mattheath: Could we prefix this with `[Client]`? Also maybe worth logging request id?

server/amqp_request.go
```diff
@@ -1,102 +1,192 @@
...
+	accessToken, _ := delivery.Headers["Access-Token"].(string)
+	parentRequestID, _ := delivery.Headers["Parent-Request-ID"].(string)
+	sessionBytes, _ := delivery.Headers["Session"].(string)
+
+	var (
+		session auth.Session
+		err     error
+	)
+	if sessionBytes != "" {
+		if s.AuthenticationProvider() != nil {
```
mattheath: Could combine these two lines into a single `if` statement?

server/rabbitmq.go
```diff
@@ -186,7 +208,8 @@ func (s *AMQPServer) respondWithError(delivery amqp.Delivery, err error) {
 	svcErr := errors.Wrap(err)
 	b, err := proto.Marshal(errors.Marshal(svcErr))
 	if err != nil {
-		// shit
+		// shit. Nothing we can do. The bytes come out empty.
+		log.Warnf("Failed to marshal error. Shit.")
```
mattheath: Sweary log lines ahoy!

mattheath:
Couple of super minor things, then think this looks good to go! :ship:

mattheath:
ci rebuild

mattheath:
ci rebuild

PR #36, opened 2015-05-19
Correctly lazily initialise rabbitmq connections

mattheath:
Otherwise we connect immediately as the default client and server are package level variables, and spew a ton of log output.


jonashuckestein:
awesome, thanks!

PR #37, opened 2015-07-01
Refactor

obeattie:
Includes changes discussed in more depth in mondough/wearedev#213. Very briefly:

* Adds the notion of a `Transport`
* Breaks servers/clients out into `mercury`
* Removes auth-specific things; these are better living in layers atop `mercury`


rabbit/utils.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func tableToHeaders(table amqp.Table) map[string]string {
+	result := make(map[string]string, len(table))
+	for k, v := range table {
+		switch v := v.(type) {
+		case string:
+			result[k] = v
+		default:
+			log.Tracef("[Bedrock:RabbitTransport] Received non-string header value for %s; discarding", k)
```
mattheath: `Bedrock` should presumably be removed here

rabbit/rabbittransport.go
```diff
@@ -0,0 +1,426 @@
...
+	return t.tomb
+}
+
+func (t *rabbitTransport) connection() *RabbitConnection {
+	t.connM.RLock()
+	defer t.connM.RUnlock()
+	return t.conn
+}
+
+func (t *rabbitTransport) Ready() <-chan struct{} {
```
mattheath: Might be worth adding a comment here to remind people (ie. me :wink:) that this returns a channel that will be closed, rather than received off. :+1: on design though.

rabbit/rabbittransport.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+func (t *rabbitTransport) Listen(serviceName string, rc chan<- message.Request) error {
+	tm := &tomb.Tomb{}
+	cn := t.deliveryChan(serviceName)
+	t.listenersM.Lock()
+	if _, ok := t.listeners[cn]; ok {
+		t.listenersM.Unlock()
+		return transport.ErrAlreadyListening
+	} else {
```
mattheath: This `else` statement isn't needed as you're returning early above?

rabbit/rabbittransport.go
```diff
@@ -0,0 +1,426 @@
...
+	return fmt.Sprintf("%s[%s]", delivery.RoutingKey, delivery.CorrelationId)
+}
+
+func (t *rabbitTransport) Respond(req message.Request, rsp message.Response) error {
+	headers := rsp.Headers()
+	headers["Content-Encoding"] = "response"
+	headers["Service"] = rsp.Service()
+	headers["Endpoint"] = rsp.Endpoint()
+
+	timeout := time.NewTimer(respondTimeout)
```
mattheath: Ideally this would be configurable at some point, and potentially even be aware of the current request deadline if/when we put these in. We can live with 10s for this PR though!

rabbit/rabbittransport.go
```diff
@@ -0,0 +1,426 @@
...
+	defer timeout.Stop()
+
+	headers := req.Headers()
+	headers["Content-Encoding"] = "request"
+	headers["Service"] = req.Service()
+	headers["Endpoint"] = req.Endpoint()
+
+	select {
+	case <-t.Ready():
+	case <-timeout.C:
```
mattheath: We're using the same timeout here for both connection and receiving a response, so this is treated more as a deadline, and could lead to us sending requests and immediately aborting listening for the responses as a result. Again, fine for now as this is a vast improvement!

rabbit/rabbittransport.go
```diff
@@ -0,0 +1,426 @@
...
+	case "request":
+		req := message.NewRequest()
+		t.deliveryToMessage(delivery, req)
+
+		timeout := time.NewTimer(chanSendTimeout)
+		defer timeout.Stop()
+		select {
+		case reqChan <- req:
+		case <-timeout.C:
+			log.Errorf("[Typhon:RabbitTransport] Could not deliver request %s after %s: receiving channel is full",
```
mattheath: In the case of a timeout pushing into the request delivery channel we drop the message here? Is that correct? In which case the corresponding client which sent this request would wait for the full timeout at the other end, rather than receiving some form of out of capacity error? Again we can deal with this later :smile:

mattheath:
I really, really, like this! :+1: :tada: :boom: :dancer: 

Couple of minor thing we could fix up before merging, but apart from that the other comments are more things we should consider separately to this, mainly as our use cases become more complex and we start tightening up timeouts and error propagation, namely:

* The timeouts for sending / receiving a response should be configurable
* In a couple of places we use the same timeout timer for both connection and receiving a response, so its more of a deadline than a timeout, this is good, though in the future we may want to pass per request deadlines through in a context, so we would need to make these configurable.
* We can potentially drop requests due to being out of capacity (good) but we don't then respond to the client, leaving the client waiting for its maximum timeout (bad). Potentially these timeouts are the same or similar, so this might not be an issue. Also with the separation of transport and server i'm not 100% certain where this should live, as it is more the responsibilty of the server, but in the case above we are unable to deliver to the server.

Otherwise, good to go! :star2:

rabbit/rabbittransport.go
obeattie: :+1: Will add a comment to the interface.

rabbit/utils.go
obeattie: Good catch

rabbit/rabbittransport.go
obeattie: Yes, we should definitely add request/response options at some point (which I guess would be best as an `Options` struct passed to the `Send` and `Respond` methods). As you say, we can probably add them when we actually want to implement configurable timeouts, though if you want me to add them now I'm totally happy to do so. :smile:

rabbit/rabbittransport.go
obeattie: Good catch, and I agree this isn't ideal. Currently, the transports are completely unaware of the request/response formats beyond the very basics, like how to retrieve the service/endpoint from a message. A "canonical" error format exists at the Mercury level, so I'm not sure adding a canonical format here as well is ideal.

I suppose a better way to deal with this might be to [negatively acknowledge](https://www.rabbitmq.com/nack.html) the message, so it can be redelivered to another instance of the service. We'd need to only retry up to a maximum number of times to avoid infinitely passing a hot potato around.

rabbit/rabbittransport.go
mattheath: Sounds great, lets add later :smile:

rabbit/rabbittransport.go
mattheath: :+1:

rabbit/rabbittransport.go
mattheath: Negatively acknowledging here sounds ideal, though I am not sure how this works as we [auto-ack messages](https://github.com/mondough/typhon/blob/46bac3bd244afd59d60d07781cb5609bf0c1137d/rabbit/channel.go#L107). Ideally this would be fine, and rabbit itself could be configured to only redeliver messages _n_ times, so that our code here remained very simple, instead of having to increment counters and actually requeue the message.

For the moment I think logging an error is Good Enough™ and we should probably punt this to some point in the future if it becomes a problem.

PR #41, opened 2015-07-23
Copy the request in a mock transport *after* its ID has been set

obeattie:
The caller can then retrieve the automatically-set request ID. This mirrors the behaviour of the Rabbit transport more closely.


mattheath:
:+1:

PR #43, opened 2015-08-17
Pass through origin information

mattheath:
Adds `OriginService` and `OriginEndpoint` concepts to typhon messages, which are marshalled on and off the wire. This allows us to pass the origin of messages which is useful for tooling like tracing.

Note these are not cryptographically signed or secure, so the server is trusting the client sent the correct information.


obeattie:
LGTM.

PR #44, opened 2015-08-18
Remove OriginService and OriginEndpoint

mattheath:
So after some major refactoring, this removes origin information added by #43 as the origin information is now represented at a higher level in [Mercury](https://github.com/mondough/mercury) middleware by mondough/mercury#9, and isn't necessary at a transport level.

For a laugh, diff this: https://github.com/mondough/typhon/compare/a6a4f48b4d05925e42b5ae277aa65d7bddd336de...removeorigin


obeattie:
Love that diff. :laughing:

PR #45, opened 2015-08-24
Add json name to Error struct

danielchatfield:


danielchatfield:
We marshal errors to JSON in the API but without this we can't use the builtin json unmarshaller.

PR #46, opened 2015-09-02
Refactor terrors

domwong:
Big refactor of errors. ***PLEASE DO NOT MERGE UNTIL MERCURY AND OTHER DEPENDENT PRS ARE ALSO READY BECAUSE THESE ARE BREAKING CHANGES***

Errors now look like this

```
type Error struct {
	Code        string            `json:"code"`
	Message     string            `json:"message"`
	Params      map[string]string `json:"params"`
	StackFrames stack.Stack       `json:"stack"`
}

```

Main changes:
* `Code` is now a string, and expected to be a dotted code. This allows us to provide more context for errors.
* `PublicContext` and `PrivateContext` are replaced with a single `Params` map. This paves the way to add `golang.org/x/net/context.Context` as service context.


errors/errors.go
```diff
@@ -47,19 +47,23 @@ func (p *Error) StackString() string {
...
-	return errorFactory(code, message, context...)
+func New(code string, message string, params map[string]string) *Error {
+	return errorFactory(code, message, params)
 }
 
 // Wrap takes any error interface and wraps it into an Error.
 // This is useful because an Error contains lots of useful goodies, like the stacktrace of the error.
 // NOTE: If `err` is already an `Error` the passed contexts will be ignored
-func Wrap(err error, context ...map[string]string) *Error {
+func Wrap(err error, params map[string]string) *Error {
```
jonashuckestein: It would be better if this method returned `error` IMO, because otherwise you can't safely call `errors.Wrap(nil)` as illustrated here:

```
func blurb() error {
  return terrors.Wrap(nil)
}

if err := blurb(); err != nil {
  // this will now execute (I think)...
}
```

Addressing this would require rewriting a bit of code in Typhon (it used to be in `endpoint.go` right after executing the endpoint handler, but not sure where it is after the mercury/typhon refactor). Essentially, if you need a `*Error` after calling `errors.Wrap`, you'll have to cast it (which you can safely do afterwards). @obeattie what do you think?

Right now we have to be careful never to do this in service code:
```
result, err := someDatabaseOperation()
return result, errors.Wrap(err) // ALARM! can only do this if err != nil
```
Stackoverflow says this is bad design anyway, but it definitely caused bugs in the past

errors/stack/stack.go
```diff
@@ -19,12 +19,12 @@ var (
 )
 
 type Frame struct {
-	Filename string `json:"filename"`
-	Method   string `json:"method"`
-	Line     int    `json:"lineno"`
+	Filename string
+	Method   string
+	Line     int
```
jonashuckestein: Why do we no longer need the json annotations? I think (not 100% sure) the RPC endpoint uses these annotations because it returns the json-marshaled error

errors/errors.go
```diff
@@ -67,79 +71,85 @@ func Wrap(err error, context ...map[string]string) *Error {
...
 }
 
-// errorConstructor returns a `*Error` with the specified code, message and context. The main work
-// consists of managing the map[string]string's at the end of the arguments list.
-// In practice we only ever pass in two: the first one is public and will be sent to the client,
-// the second one is private and can contain internal information that is useful for debugging
-func errorFactory(code int, message string, context ...map[string]string) *Error {
+// errorConstructor returns a `*Error` with the specified code, message and params.
+// Builds a stack based on the current call stack
+func errorFactory(code string, message string, params map[string]string) *Error {
```
jonashuckestein: Not sure if this is really bad design, but I would have made `params` optional so that we don't have to put `nil` at the end of the constructor every time we use it. FWIW `Sprintf` uses a similar trick

jonashuckestein:
A few minor comments, but otherwise looks really good :) I'm don't care deeply about any of the points I raised

errors/errors.go
domwong: Indeed. Something like this illustrates the point http://play.golang.org/p/5aus8SlMdF

I think returning error rather than Error is the recommendation (can't remember where I read that now)

domwong: http://golang.org/doc/faq#nil_error

"It's a good idea for functions that return errors always to use the error type in their signature (as we did above) rather than a concrete type such as *MyError, to help guarantee the error is created correctly. As an example, os.Open returns an error even though, if not nil, it's always of concrete type *os.PathError."

errors/errors.go
domwong: Yeah I'm in two minds about this. On the one hand using the variadic operator makes it easier (no nils) but on the other hand it diminishes the meaning of it. i.e. you can legitimately pass multiple maps but what does that supposed to do?

errors/stack/stack.go
danielchatfield: Can confirm, the RPC endpoints use the annotations.

domwong: Delete error from previous commit. Will reinstate. Good catch

errors/errors.go
```diff
@@ -47,19 +47,27 @@ func (p *Error) StackString() string {
...
 // Otherwise use the handy shorthand factories below
-func New(code int, message string, context ...map[string]string) *Error {
-	return errorFactory(code, message, context...)
+func New(code string, message string, params map[string]string) *Error {
+	return errorFactory(code, message, params)
 }
 
 // Wrap takes any error interface and wraps it into an Error.
 // This is useful because an Error contains lots of useful goodies, like the stacktrace of the error.
 // NOTE: If `err` is already an `Error` the passed contexts will be ignored
```
mattheath: mentions `context` but should be `params`.

Alternatively, could we merge these into the existing error in the case the map is populated?

errors/errors.go
mattheath: Given its an unexported method making `params` optional doesn't matter outside of this package? If we did make it variadic then we should attempt to merge the multiple maps with the ordering clearly documented.

errors/errors.go
```diff
@@ -2,64 +2,73 @@ package errors
...
 		stackStr = fmt.Sprintf("%s\n  %s:%d in %s", stackStr, frame.Filename, frame.Line, frame.Method)
 	}
 	return stackStr
 }
 
-// VerboseString returns the error message, stack trace and contexts
+// VerboseString returns the error message, stack trace and params
 func (p *Error) VerboseString() string {
-	return fmt.Sprintf("%s\nPublic context: %+v\nPrivate context: %+v\n%s", p.Error(), p.PublicContext, p.PrivateContext, p.StackString())
+	return fmt.Sprintf("%s. %s\nParams: %+v\n%s", p.Error(), p.Message, p.Params, p.StackString())
```
obeattie: If I'm reading this right, the output will be `dotted.code: message. message\nstack`

errors/errors.go
```diff
@@ -2,64 +2,73 @@ package errors
...
-// Error returns a string message of the error
+// Error returns a string message of the error. It is a concatenation of Code and Message params
 // This means the Error implements the error interface
 func (p *Error) Error() string {
 	if p == nil {
 		return ""
 	}
-	return p.Message
+
+	return fmt.Sprintf("%s: %s", p.Code, p.Message)
```
obeattie: If `Message` is blank this would output `"dotted.code: "` whereas we probably want it just to return `"dotted.code"`. Minor comment though.

errors/errors.go
```diff
@@ -67,79 +76,92 @@ func Wrap(err error, context ...map[string]string) *Error {
...
+		return code
+	}
+	return strings.Join([]string{prefix, code}, ".")
+}
+
+// Matches returns whether the string returned from error.Error() contains the given param string. This means you can
+// match the error on different levels e.g. dotted codes `bad_request` or `bad_request.missing_param` or even on the
+// more descriptive message
+func (p *Error) Matches(match string) bool {
+	return strings.Contains(p.Error(), match)
```
obeattie: Would we want this to be a prefix match, rather than an "anywhere in the string" match? (ie. [`strings.HasPrefix`](http://golang.org/pkg/strings/#HasPrefix)?)

errors/errors.go
domwong: good spot

proto/error/error.proto
```diff
@@ -7,9 +7,11 @@ message StackFrame {
 }
 
 message Error {
-	int32 code = 1;
+	//int32 deprecated_code = 1;
```
obeattie: Can we remove these rather than commenting them? Also, given this is a breaking change to all the things anyway, I'd just number the proto fields sequentially.

obeattie:
A few (pretty minor) points. Otherwise LGTM :rocket:.

errors/errors.go
domwong: Good question. I like it being contains because you can then match on the old style message too. For example if you wrap a traditional error, otherwise you'd have to match on `"internal_service: foo bar message"`

obeattie: Fair point. If it means we'd have to rewrite many many more things for questionable benefit, it's probably not worth doing the prefix-only match.

PR #47, opened 2015-09-28
Recover *Error from any error

mattheath:
Helper method to recover a typhon `*Error` from any error in a similar way to `Wrap`, setting the code to `ErrInternalService` if this error wasn't already an `*Error`. This is mainly to reduce casting in my other code. Thoughts?

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -87,6 +87,21 @@ func WrapWithCode(err error, params map[string]string, code string) error {
 	}
 }
 
+// Recover returns a typhon Error from any error interface
+// If `err` is not already a typhon `Error` then it will be wrapped as an
+// Internal Service error
+func Recover(err error) *Error {
+	if err == nil {
+		return nil
+	}
+	switch err := err.(type) {
+	case *Error:
+		return err
+	default:
+		return errorFactory(ErrInternalService, err.Error(), nil)
+	}
+}
+
 // InternalService creates a new error to represent an internal service error.
 // Only use internal service error if we know very little about the error. Most
 // internal service errors will come from `Wrap`ing a vanilla `error` interface
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -73,6 +73,8 @@ func Wrap(err error, params map[string]string) error {
 	return WrapWithCode(err, params, ErrInternalService)
 }
 
+// WrapWithCode takes any error interface and wraps it into an Error.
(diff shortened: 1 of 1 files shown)


domwong:
I think this looks reasonable, only thing I wonder about is the naming since recover has special meaning in golang but I guess it's still clear what it's doing so shouldn't cause too much confusion.

@obeattie thoughts?

mattheath:
Good point - yeah, open to suggestions :smile:

obeattie:
I'm confused what the purpose of this is. These to me both seem equivalent:

```go
terrors.Wrap(err, nil)
terrors.Recover(err)
```

Is the point to eliminate needing to pass `nil`? If so, we could (possibly) change the `Wrap` signature to

```go
func Wrap(err error params ...map[string]string)
```

obeattie:
Oh, I see. The crucial part is that one returns `error` whereas the other returns `*Error`. Perhaps `Wrap` should return `*Error`?

obeattie:
Oh, I see. The crucial part is that one returns `error` whereas the other returns `*Error`. Perhaps `Wrap` should return `*Error`? I'd say we should either do that, or say that calling code should just do:

```go
terrors.Wrap(err, nil).(*terrors.Error)
```

This seems to be compliant with "conventions" from the Go standard library. For instance, `http.ResponseWriter` can be type-asserted to `http.Flusher` and `http.CloseNotifier`, and they document this.

On one hand, type assertions are fairly straightforward. On the other, they can be confusing when reading code since they require extra context to understand.

Not really sure what the best approach is here, but I think having both `Wrap` and `Recover` is confusing.

mattheath:
I thought `Wrap` deliberately returned the `error` interface (and an explicit `nil`) to get around problems with a `nil` error? https://golang.org/doc/faq#nil_error

This was mainly to avoid typing something like this in certain places:
```go
if err != nil {
	switch terr := err.(type) {
	case *terrors.Error:
		// use the code & message directly
		return fmt.Sprintf("<some xml cruft><code>%s</code><reason>%s</reason></xml>", terr.Code, terr.Message)
	default:
		// make up a code, and use the error interface to get a message
	}
}
```

Though as you point out, re-wrapping a non-nil error then type asserting is neater and should be safe in all cases?

domwong:
Yep that's why we did as `error` rather than `*Error`. Using `Wrap` with type assertion on non nil errors might be the happy medium then

mattheath:
Cool, will do that :+1:

PR #48, opened 2015-10-11
Use direct reply-to instead of long-lived dedicated queue

boyand:
This replaces the dedicated long lived reply queue with direct reply-to as described here: https://www.rabbitmq.com/direct-reply-to.html


obeattie:
This is great; thank you :boom:

PR #49, opened 2015-10-12
Make Service and Endpoint request-only fields

obeattie:
Decided that they don't make any sense on a `Response`… they were ambiguous at best. :bomb:


mattheath:
Agreed, it certainly wasn't obvious what these represented on a response :+1:

PR #54, opened 2016-04-14
Catch two unchecked interface conversions

mattheath:
Catch two places where an interface conversion wasn't checked for success, which triggers an unrecovered panic (and therefore service exit). :boom: 

Reproducible by sending a request with a non-string `"Content-Encoding"` header to the channel the transport listens on, which triggers the panic. :fire:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: go
 go:
   - 1.5.2
   - 1.6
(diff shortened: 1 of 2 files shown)


obeattie:
LGTM :+1:
Really good catch :fishing_pole_and_fish:

PR #56, opened 2016-05-27
HTTP

obeattie:
This PR is practically a rewrite of Typhon.


toqueteos:
You are missing updating the README now that Rabbit is not supported/required anymore.

request.go
```diff
@@ -0,0 +1,110 @@
...
+	"golang.org/x/net/context"
+)
+
+type Request struct {
+	http.Request
+	context.Context
+}
+
+// Encode serialises the passed object as JSON into the body (and sets appropriate headers).
+func (r *Request) Encode(v interface{}) {
```
mattheath: I guess at some point it would be nice to add support for alternative encoders, but we can still use this interface at that point anyway 😄 

The `Decode` method below returns an `error`, should this not too? Otherwise we presumably can only determine encoding failures from logs, and possibly weird errors as the `Content-Type` header & body not being set. I'm guessing that the `Request` type not having an `Error` param (unlike `Response`) and `NewRequest` not returning an error makes this more difficult, hence [this](https://github.com/mondough/typhon/blob/5facb016801ef1d2cf81dbcac5cdb44f703b6bef/request.go#L101), but given it would be a single-value return we could still return the error here?

response.go
```diff
@@ -0,0 +1,110 @@
...
+	if r.Response == nil {
+		err = terrors.InternalService("", "Response has no body", nil)
+	} else {
+		err := json.NewDecoder(r.Body).Decode(v)
+		if err != nil {
+			log.Warn(r.ctx, "Failed to decode response body: %v", err)
+		}
+		err = terrors.WrapWithCode(err, nil, terrors.ErrBadResponse)
+	}
+	if r.Error == nil {
```
mattheath: In the case this is non-nil, but `r.Response` is nil we throw away the error without logging this.

request.go
```diff
@@ -0,0 +1,110 @@
...
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	httpReq, _ := http.NewRequest(method, url, nil) // @TODO: Don't swallow this error
+	httpReq.Body = &bufCloser{}
+	req := Request{
+		Request: *httpReq,
+		Context: ctx}
+	if body != nil {
+		req.Encode(body)
```
mattheath: We're also swallowing the error here, in addition to the `http.NewRequest` a couple of lines above. Could we not return `(Request, error)` from the `NewRequest` method bringing this in-line with `http.NewRequest` in the stdlib? Is there anywhere we would use `NewRequest` in a single-value return context?

response.go
obeattie: In that case, the error would be returned to the caller I think?

request.go
obeattie: I'd prefer not to have to return an error here, otherwise everywhere requests are made, there needs to be two lots of error checking:

```go
req, err := typhon.NewRequest(ctx, "POST", "/", nil)
if err != nil {
    // error stuff
}
rsp := req.Send().Response()
if rsp.Error != nil {
    // error stuff
}
```

That amount of boilerplate just to make a request is really tedious. I think a better approach might be to include a hidden `error` member inside the `Request` struct, that the `Client` looks for when sending the request. If the request contains an error, it returns an error response without ever dispatching the request. This is similar to the old behaviour in Mercury. What do you think?

router.go
```diff
@@ -0,0 +1,108 @@
...
+
+type Router interface {
+	// OPTIONS is a shortcut for Register("PUT", svc).
+	OPTIONS(path string, svc Service)
+	// GET is a shortcut for Register("GET", svc).
+	GET(path string, svc Service)
+	// HEAD is a shortcut for Register("HEAD", svc).
+	HEAD(path string, svc Service)
+	// POST is a shortcut for Register("Delete", svc).
+	POST(path string, svc Service)
```
mattheath: Register("POST")

router.go
```diff
@@ -0,0 +1,108 @@
...
+	// PUT is a shortcut for Register("Delete", svc).
```
mattheath: `Register("PUT")`

router.go
```diff
@@ -0,0 +1,108 @@
...
+	PUT(path string, svc Service)
+	// DELETE is a shortcut for Register("Delete", svc).
```
mattheath: `Register("DELETE")` sorry, capitalisation 😉

router.go
```diff
@@ -0,0 +1,108 @@
...
+	DELETE(path string, svc Service)
+	// TRACE is a shortcut for Register("Delete", svc).
```
mattheath: `Register("TRACE")` 😉

router.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/mondough/terrors"
+)
+
+type Router interface {
+	// OPTIONS is a shortcut for Register("PUT", svc).
```
mattheath: `Register("OPTIONS"`

router.go
```diff
@@ -0,0 +1,108 @@
...
+func (r *router) HEAD(path string, svc Service)    { r.Register("HEAD", path, svc) }
+func (r *router) POST(path string, svc Service)    { r.Register("POST", path, svc) }
+func (r *router) PUT(path string, svc Service)     { r.Register("PUT", path, svc) }
+func (r *router) DELETE(path string, svc Service)  { r.Register("DELETE", path, svc) }
+func (r *router) TRACE(path string, svc Service)   { r.Register("TRACE", path, svc) }
+
+// I'm sorry, dear reader, I really am. To do this properly is more work than I have the appetite for right now.
+//
+// Future me will remove this horrific cruft and provide a URL router that acts on Services directly, without needing
+// the kabuki of a fake Handler and ResponseWriter.
```
mattheath: ❤️

response.go
mattheath: Sorry, I meant we throw away the _"Response has no body"_ error, and return the previous error. So the caller is return _an_ error, which is probably fine, and preferable in this case, but given we log the one above (line 38), we may also want to log this?

request.go
mattheath: Yep, really like this approach. Keeps the interface clean, and I've seen a few other examples like this before when chaining. 👍

mattheath:
Love this ❤️ 

Couple of super minor things and we're good to go! 🚀

mattheath:
❤️ 🐰 🔫

PR #57, opened 2016-08-09
Ensure to cancel contexts

obeattie:
This looks to be the source of a goroutine bug. Where control escapes the containing functions here, there's a good chance cancel() may have already been called – if so, this is a no-op.


jonashuckestein:
LGTM 🚀

PR #70, opened 2017-01-29
Switch out router internals

obeattie:
This resolves a common frustration where two "conflicting" paths like `{/, /:foo}` cannot be registered.


dancannon:
Looks great 👍 Ill try and update the routing for monzo.me soon as well.

PR #76, opened 2017-10-11
travis: update go version

dvrkps:


obeattie:
👍 Thanks for this. Agree with testing using the latest version, but it would be good to _also_ test using `1.8` and `1.9`. Could you add those two please?

dvrkps:
@obeattie done.

PR #77, opened 2017-12-10
Allow streaming responses 🏞

obeattie:
This allows services to stream responses back to clients, and clients to process such responses gradually. If a service wishes to return a streaming response, it sets `response.Body` to a new `Steamer` object to which it can then write. This interface does not provide buffering; if services need this they can layer a buffered (`bufio`) writer on top.

This changes the client's internal behaviour quite significantly, but this should not affect the API exposed to users. Specifically, a client would previously fully download a response and read it into an in-memory buffer before `HttpService` (used most commonly inside `request.Send()`) would return. Now, this is not the case; `HttpService` returns as soon as it can, and before `response.Body` is fully read.

The previous behaviour mostly existed to defend against callers which did not `Close()` response bodies, which is a common source of resource leaks in Go programs. We have the advantage that we propagate contexts throughout the (distributed) call stack, and Typhon ensures that a context is cancelled when a *server* has fully written a response to a client. This means we can rely on this behaviour even in the absence of buffering to provide a guarantee that resources will be released. Consider this service:

 ```go
func Dummy(req typhon.Request) typhon.Response {
    typhon.NewRequest(nil, "GET", "http://www.google.com/", nil).Send()
    return req.Response("haha")
}
```

Even though it uselessly sends a request to `http://www.google.com/` and does nothing to ensure the response body is `Close()`d, as soon as the `Dummy` handler returns its response to its upstream, the context used to call Google is cancelled. This cancellation is propagated down to the client, which then ensures the body is `Close()`d.

The new approach also has the advantage of not unnecessarily buffering responses in-memory.

### CODE REVIEW

File: chunk.go [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+			}
+			if flusherOk {
+				flusher.Flush()
+			}
+			if nr != nw {
+				err = io.ErrShortWrite
+				break
+			}
+		}
+		if er != nil {
```
mattheath: Almost caught me here with `er`, but all good 😄

chunk.go
```diff
@@ -0,0 +1,43 @@
...
+		if nr > 0 {
+			nw, ew := dst.Write(buf[0:nr])
+			if nw > 0 {
+				written += int64(nw)
+			}
+			if ew != nil {
+				err = ew
+				break
+			}
+			if flusherOk {
```
mattheath: Would this always be `true`? We check `!flusherOK` above and return, and I can't see a way this can change?

obeattie: I added this before I added the `!flusherOK` check, so you're right this is redundant now. Will change 👍

chunk.go
obeattie: Yeah, I copied and pasted this directly from the source of the `io` package as the comment above mentions. I considered changing this too but thought it would probably be easier to maintain if I changed as little as possible. 🤔

chunk.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func copyChunked(dst io.Writer, src io.Reader) (written int64, err error) {
+	flusher, flusherOk := dst.(http.Flusher)
+	if !flusherOk {
+		return io.Copy(dst, src)
+	}
+
+	// This is taken and lightly adapted from the source of io.Copy
+	buf := make([]byte, 32*1024)
```
andreishappy: This may be a very ignorant question, but I'd like to know. Why is this 32*1024?

obeattie: I'm honestly not sure. I copied this code directly from the `io` package. I just went and looked up the history and I can see that this code dates [all the way back to 2008](https://github.com/golang/go/commit/79d94d504f8f3e82e994a4f63d37f56cebc6e7cc). Sadly the commit message doesn't provide any information as to _why_ this number was chosen, but I suspect it's a somewhat arbitrary choice that also is a multiple of the default page size in Linux and most other operating systems.

README.md
```diff
@@ -1,6 +1,13 @@
...
 [![Build Status](https://travis-ci.org/monzo/typhon.svg?branch=master)](https://travis-ci.org/monzo/typhon)
 [![GoDoc](https://godoc.org/github.com/monzo/typhon?status.svg)](https://godoc.org/github.com/monzo/typhon)
 
-A transport layer for micro-service communication.
+Typhon is a thin wrapper around `net/http` that we use at Monzo to build RPC servers and clients in our
+microservices platform.
+
+It provides a number of conveniences for doing things like injecting middleware "filters", encoding and decoding
+responses, response streaming, propagating cancellation, passing errors. Its API is deliberately constrained but
+intended to promote safety: for example, clients are freed from the worry of leaking resources if they fail call
```
kjgorman: grammar: "fail call" -> "fail to call" 🤓

PR #79, opened 2018-03-05
Don't chunk non-streaming request bodies 🎁

obeattie:
For small requests it is not desirable to chunk the body as it means intermediary proxies (like linkerd) cannot retry the request transparently.

Go's logic for deciding when to chunk a request is subtle and a bit [magic](https://github.com/golang/go/blob/26708439ecb4dc923149b17eb522df46dcc2fb23/src/net/http/transfer.go#L150-L186), but this prevents chunking behaviour when `Encode()` is called before sending the response off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - 1.8.x
   - 1.9.x
+  - 1.10.x
 
 install:
   - export PATH=${PATH}:${HOME}/gopath/bin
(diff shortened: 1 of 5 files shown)


obeattie:
Yes, sorry, that is sloppy wording on my part. The decision to chunk or not has nothing to do with the request size at this level, but instead is a function of whether the body is fully-buffered before being sent.

Separately, we may want to introduce such a size threshold, but I think doing so is outside the scope of this PR.

mattheath:
Perfect, thanks! 😄

PR #80, opened 2018-03-07
Fix up README file

obeattie:
…and add an `.editorconfig` to prevent things like trailing spaces being stripped in Markdown files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+[*]
+charset = utf-8
+end_of_line = LF
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+[*]
```
TransactCharlie: Should .editorconfig be part of pr? maybe it should be in .gitignore?

evnsio: I believe adding this is the purpose of the MR?  We want the `.editorconfig` in the repo to fix the mis-formatting of .md files.

PR #81, opened 2018-03-15
Fix leak when infinite contexts are used to make requests 🚰

obeattie:
Previously when a caller used an infinite context (eg. `context.Background()`) to make a request, this would cause a resource leak as Typhon was waiting for the context to be cancelled to `Close()` the body.

This change ensures request bodies are additionally closed when they are fully read, and adds tests to ensure no goroutines are leaked.

### DIFF
--- a/client.go
+++ b/client.go
@@ -50,10 +50,14 @@ func HttpService(rt http.RoundTripper) Service {
 		// When the calling context is cancelled, close the response body
 		// This protects callers that forget to call Close(), or those which proxy responses upstream
 		if httpRsp != nil && httpRsp.Body != nil {
-			body := httpRsp.Body
+			dr := newDoneReader(httpRsp.Body)
+			httpRsp.Body = dr
 			go func() {
-				<-req.Done()
-				body.Close()
+				select {
+				case <-dr.done:
+				case <-req.Done():
+					dr.Close()
+				}
 			}()
 		}
 		return Response{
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -50,10 +50,14 @@ func HttpService(rt http.RoundTripper) Service {
...
 		// This protects callers that forget to call Close(), or those which proxy responses upstream
 		if httpRsp != nil && httpRsp.Body != nil {
-			body := httpRsp.Body
+			dr := newDoneReader(httpRsp.Body)
+			httpRsp.Body = dr
 			go func() {
-				<-req.Done()
-				body.Close()
+				select {
+				case <-dr.done:
```
priyeshpatel: Do we _also_ need to close the body in the first case too (whilst ensuring that we only close it once)? If we incur an `io.EOF` whilst reading, then we will close the done channel but the body won't be closed.

client.go
```diff
@@ -50,9 +50,13 @@ func HttpService(rt http.RoundTripper) Service {
 		// When the calling context is cancelled, close the response body
 		// This protects callers that forget to call Close(), or those which proxy responses upstream
 		if httpRsp != nil && httpRsp.Body != nil {
-			body := httpRsp.Body
+			body := newDoneReader(httpRsp.Body)
+			httpRsp.Body = body
 			go func() {
-				<-req.Done()
+				select {
```
robbiev: I'm wondering if the tests using the leak detector could be flaky as closing happens in a separate goroutine here?

obeattie: The leak detector waits for up to a second by default for goroutines to close by default. Also, the tests definitely fail on the old version of the code (I updated the tests before writing the code to fix this issue), so I'm fairly confident it's working.

PR #82, opened 2018-03-19
Context handling

obeattie:
This contains various improvements to context handling, with the goal of eliminating context goroutine leaks. Specifically:

* Request cancellation behaviour can be controlled by using only context cancellation. To eliminate some complexity, remove the cancellation hooks from `ResponseFuture` (which are not used in Monzo's code anyway)
* The context library contains some performance optimisations when dealing with its own context types. For users of Typhon, it is common to use a Request as a context, preventing these performance optimisations from working properly. Internally, Typhon will now unwrap these contexts.
* If a context can never be cancelled, it returns a `nil` channel on `.Done()`. In this case, don't kick off a background goroutine that responds to cancellation in the client.

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -76,7 +76,7 @@ func (r *doneReader) Close() error {
 func (r *doneReader) Read(p []byte) (int, error) {
 	n, err := r.ReadCloser.Read(p)
 	if err == io.EOF {
-		r.closedOnce.Do(func() { close(r.closed) })
+		r.Close()
 	}
 	return n, err
 }
--- a/buffer.go
+++ b/buffer.go
@@ -56,27 +56,27 @@ func (w *countingWriter) Write(p []byte) (int, error) {
 // doneReader is a wrapper around a ReadCloser which provides notification when the stream has been fully consumed
 // (ie. when EOF is reached, or when the reader is closed.)
 type doneReader struct {
-	done     chan struct{}
-	doneOnce sync.Once
+	closed     chan struct{}
+	closedOnce sync.Once
 	io.ReadCloser
 }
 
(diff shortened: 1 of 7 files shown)


buffer.go
```diff
@@ -56,27 +56,27 @@ func (w *countingWriter) Write(p []byte) (int, error) {
...
-	r.doneOnce.Do(func() { close(r.done) })
+	r.closedOnce.Do(func() { close(r.closed) })
 	return err
 }
 
 func (r *doneReader) Read(p []byte) (int, error) {
 	n, err := r.ReadCloser.Read(p)
 	if err == io.EOF {
-		r.doneOnce.Do(func() { close(r.done) })
+		r.closedOnce.Do(func() { close(r.closed) })
```
priyeshpatel: I note that we've renamed the channel from `done` to `closed`, and that we're no longer closing the `body` in the `<-body.closed` case in `client.go`. Does this imply that we also need to close the body in this `if` statement prior to signalling on the `closed` channel too?

client.go
```diff
@@ -23,67 +22,69 @@ var (
...
 	return f.done
 }
 
+// Response provides access to the response object, blocking until it is available
 func (f *ResponseFuture) Response() Response {
 	<-f.WaitC()
 	return f.r
 }
 
-func (f *ResponseFuture) Cancel() {
```
TransactCharlie: This is the only signature change which is not obvious (to me) if anyone will ever use this directly. We do *return* a ResonseFuture from SendVia (which is a public function) but whether a third party client would have code that attempts to call Cancel() isn't obvious.

Do we care or need to about breaking any protocol here? should this be a noop instead but still be supported?

PR #88, opened 2018-06-26
Expose HTTP server to allow clients to override config

danielchatfield:
It is sometimes desirable to alter the configuration of the `*http.Server` when running a typhon service. An example use case is TLS. TLS can also be done by calling `typhon.Serve` with a TLS connection directly but if doing this the `TLS` property on the http request will be `nil` as the http server won't be able to access properties of the connection. This is undesirable for use cases that require inspecting the client certificate.

Prior to this change typhon would create the http server and turn it into a typhon server and not give the caller any opportunity to do anything to it. Typhon already exposed `HttpServer` but didn't expose anything that let you turn that into a typhon server.

This PR introduces `ServeHTTPServer` to let a caller that require more control compose `HttpServer` and `ServeHTTPServer` themselves, reconfiguring the http server as they see fit.

### DIFF
--- a/server.go
+++ b/server.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/facebookgo/httpdown"
 	log "github.com/monzo/slog"
+	"net/http"
 )
 
 const DefaultListenAddr = "127.0.0.1:0"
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -73,3 +67,17 @@ func Listen(svc Service, addr string) (Server, error) {
 	}
 	return Serve(svc, l)
 }
+
+// ServeHTTPServer returns a typhon server for the given HTTP server.
+// This supports callers that that need to alter the configuration of
+// the HTTP server (e.g. to support TLS).
+func ServeHTTPServer(s *http.Server, l net.Listener) (Server, error) {
```
danielchatfield: @obeattie  Should I:
 - Name this `ServeHttpServer` to match `HttpServer`
 - Rename `HttpServer` to `HTTPServer` (technically a breaking change)
 - Name this `ServeHTTPServer` even though it is inconsistent

PR #90, opened 2018-09-13
Assorted minor fixes and improvements 🍱

obeattie:
• If `http.NewRequest` returned an error (incredibly rare) when constructing a new `typhon.Request`, it was silently swallowed. Now it's stored within the request object to be returned when the request is used, as we do for encoding errors.
• Test subtle behaviour when reading response bodies.
• Benchmark for full request/response cycle.
• Slight performance optimisation when writing responses out to the wire: only call the getter for the header map once rather than `len(headers)` times.


e2e_test.go
```diff
@@ -372,3 +372,34 @@ func (suite *e2eSuite) TestResponseAutoChunking() {
...
+		rsp := req.Response(nil)
+		rsp.Header.Set("a", "b")
+		rsp.Header.Set("b", "b")
+		rsp.Header.Set("c", "b")
+		return rsp
+	})
+	addr := &net.UnixAddr{
+		Net:  "unix",
+		Name: "/tmp/typhon-test.sock"}
+	l, _ := net.ListenUnix("unix", addr)
```
jaseemabid: @obeattie TIL you can listen on a unix socket instead of network address and the rest of the library works the same! 💡 Will there be any difference in benchmarks because of the change in socket type? Is there a case in which this benchmark reports a better number while the actual RTT gets worse? Just curious.

response.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/monzo/terrors"
 )
 
+// A Request is Typhon's wrapper around http.Response, used by both clients and servers.
```
jaseemabid:
```
15   1 warning  golint comment on exported type Response should be of the form "Response ..." (with optional leading article) (gometalinter)
```

obeattie: Ah shit, yes, that's a typo. I'll fix this.

PR #101, opened 2018-10-28
HTTP/2.0 🦄

obeattie:
This PR adds full HTTP/2.0 support to Typhon. There are some fairly big implementation changes to make this possible, but the API is pretty much unchanged.

* **[h2c](https://httpwg.org/specs/rfc7540.html#discover-http) support** 🆙
 Typhon servers that include the `H2cFilter` support h2c. That is, the server's functionality is available via _both_ HTTP/1.1 and HTTP/2.0 on the same listener, with clients being able to "upgrade" their connection to HTTP/2.0.
  
  Because connection draining is important for Monzo's microservices, and the Go standard library does not support it (golang/go#26682), the implementation is fairly ugly.
* **Connection hijacking** 🧛‍♀️
  As is the case with the Go standard library, it's now possible to hijack the underlying connection from within a Typhon `Service`. `Response.Writer()` can be type-asserted to a `net/http.Hijacker` to do this. This is used internally in the h2c implementation.
* **Performance improvements** 🏎 
  Server performance is improved fairly substantially by pooling buffers that are used internally when copying response bodies. Fewer allocations means less GC pressure.

All of the end-to-end tests now run against the four HTTP "flavours": `http1.1` (vanilla HTTP/1.1), `http1.1-tls` (HTTP/1.1 over TLS), `http2.0-h2c` (HTTP/2.0 with h2c upgrade) and `http2.0-h2` (HTTP/2.0 over TLS).

----

**There are lots of noisy changes to `vendor/` in this PR.** Here is the diff excluding those: https://github.com/monzo/typhon/pull/101/files/f6c74e1e3fb5991e8f286ce5625af24a0a681231


http.go
```diff
@@ -16,23 +19,40 @@ const (
 	chunkThreshold = 5 * 1000000 // 5 megabytes
 )
 
+var httpChunkBufPool = sync.Pool{
```
obeattie: This is a performance optimisation and _hugely_ reduces memory allocations. When running `BenchmarkRequestResponse`, it has shaved off >600MB of allocations.

client.go
```diff
@@ -13,13 +12,12 @@ var (
...
 	Client Service = BareClient
 	// RoundTripper is used by default in Typhon clients
-	RoundTripper http.RoundTripper = &httpcontrol.Transport{
+	RoundTripper http.RoundTripper = &http.Transport{
 		Proxy:               http.ProxyFromEnvironment,
 		DisableKeepAlives:   false,
 		DisableCompression:  false,
-		MaxIdleConnsPerHost: 10,
-		DialKeepAlive:       10 * time.Minute,
-		MaxTries:            6}
```
suhailpatel: This is interesting, I remember noticing this when investigating retries and forgot to bring this up. 

Just so I understand this clearly, for `GET` requests did we potentially have multiple things retrying (eg if Linkerd does retries and it gives up)

h2c.go
```diff
@@ -0,0 +1,145 @@
...
+		return rsp
+	}
+	return svc(req)
+}
+
+// Dear reader: I'm sorry, the code below isn't fun. This is because Go's h2c implementation doesn't have support for
+// connection draining, and all the hooks that would make this easy are unexported.
+//
+// If this ticket gets resolved this code can be dramatically simplified, but it is not a priority for the Go team:
+// https://github.com/golang/go/issues/26682
```
suhailpatel: I just read through this ticket, slightly unfortunate this is the case

response.go
```diff
@@ -14,8 +14,9 @@ import (
 // A Response is Typhon's wrapper around http.Response, used by both clients and servers.
 type Response struct {
 	*http.Response
-	Error   error
-	Request *Request // The Request that we are responding to
+	Error    error
+	Request  *Request // The Request that we are responding to
```
suhailpatel: `Request` is also exported as part of `*http.Response`. It isn't immediately clear to me why we need this to be `*typhon.Request` here?

client.go
obeattie: I _believe_ that this only applied to `net.Dial` failures – ie. a network connection to the destination couldn't be established – not application failures which result in a `5xx` response.

response.go
obeattie: The `net/http` documentation [states](https://golang.org/pkg/net/http/#Response):
>This is only populated for Client requests.

Typhon populates this field regardless of whether it is handling the request as a client or a server. Also, users of Typhon find it more convenient to use its wrapper types.

PR #105, opened 2018-11-23
Expose error in typhon.Request 🙈

suhailpatel:
Those using `typhon.NewRequest` directly have no way of reading any underlying errors generated at request creation time since no error is returned explicitly.

This PR changes `typhon.Request` to make the `err` public as `Error`

### DIFF
--- a/request.go
+++ b/request.go
@@ -15,7 +15,7 @@ import (
 type Request struct {
 	http.Request
 	context.Context
-	err error // Any error from request construction; read by Client
+	Error error // Any error from request construction
 }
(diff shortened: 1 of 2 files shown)


obeattie:
Hm I would prefer if we made this kind of panic go away without needing people to check the `.Error` property (in this case, by ensuring there is a valid – though useless – nested request.)

suhailpatel:
I am closing this PR because a lot of this code has been refactored under the hood in some other PRs and I believe this no longer applies

PR #107, opened 2018-12-17
Time out if reading headers takes more than 10 seconds

jackkleeman:


obeattie:
The `HttpServer` function has been removed in #108.

PR #111, opened 2018-12-31
Add simple service example

JoshJ1091:
This change adds an example based on go-kit's stringsvc1 example. I don't know how your services are organized, so I used a generic project format. The business logic is separated from the transport layer, and there is a client for calling the service.

I hope this might be useful for someone.

Thanks,
Josh

### DIFF
--- a/examples/stringsvc/cmd/client/main.go
+++ b/examples/stringsvc/cmd/client/main.go
@@ -1,0 +1,23 @@
+package main
+
+import (
+	"github.com/monzo/typhon"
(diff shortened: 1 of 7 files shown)


obeattie:
Sorry for taking so long to respond to this and thanks for the PR!

While I'd love to include more examples in the repo, this does not seem very minimal. I would prefer if any examples focussed on direct use the library rather than trying to promote conventions for the structure of calling code. What you have here is _very_ different to how we use Typhon within Monzo, for example.

Does this make sense?

obeattie:
Please feel free re-open if you are interested in refining the example 😄

PR #113, opened 2019-03-08
Set content length in the request/response Write method

jackkleeman:
Try to set content length correctly when using `Write()` and `Writer()` methods; only use chunked encoding with content length of -1 when the length is large, or when the handler specifically sets the body to a streamer or another io.Reader. Since the HTTP2 changes were merged, `rsp.Write([]byte("hello"))` would have content length set at -1, and chunked encoding would be used. This led to regressions where callers weren't happy with chunked encoding.

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -128,7 +128,7 @@ func TestE2E(t *testing.T) {
 			"b": "a"}, body)
 		// The response is simple too; shouldn't be chunked
 		assert.NotContains(t, rsp.TransferEncoding, "chunked")
-		assert.Equal(t, int64(10), rsp.ContentLength)
+		assert.EqualValues(t, 10, rsp.ContentLength)
 	})
 }
 
@@ -328,7 +328,7 @@ func TestE2ENoFollowRedirect(t *testing.T) {
 		rsp := req.Send().Response()
 		assert.NoError(t, rsp.Error)
 		assert.Equal(t, http.StatusFound, rsp.StatusCode)
-		assert.Equal(t, int64(56), rsp.ContentLength)
+		assert.EqualValues(t, 56, rsp.ContentLength)
 	})
 }
 
@@ -369,7 +369,7 @@ func TestE2EProxiedStreamer(t *testing.T) {
 		if !rsp.ProtoAtLeast(2, 0) {
 			assert.Contains(t, rsp.TransferEncoding, "chunked")
 		}
-		assert.True(t, rsp.ContentLength < 0)
+		assert.EqualValues(t,-1, rsp.ContentLength)
 		for i := 0; i < 100; i++ {
 			chunks <- true
 			b := make([]byte, 500)
@@ -458,7 +458,7 @@ func TestE2ERequestAutoChunking(t *testing.T) {
 		rsp = req.Send().Response()
 		require.NoError(t, rsp.Error)
 		assert.Equal(t, http.StatusOK, rsp.StatusCode)
-		assert.Equal(t, rsp.ContentLength, int64(5))
+		assert.EqualValues(t, 5, rsp.ContentLength)
 		assert.False(t, receivedChunked)
 
 		// Large request using Encode(); should be chunked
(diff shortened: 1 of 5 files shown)


response.go
```diff
@@ -96,8 +99,22 @@ func (r *Response) Write(b []byte) (int, error) {
 			rc.Close()
 		}
 		r.Body = buf
-		return buf.Write(b)
+		cw = &countingWriter{
+			Writer: buf}
+	}
+
+	n, err := cw.Write(b)
+	if r.ContentLength < 0 && cw.n < chunkThreshold {
```
obeattie: I think you should only do this if there was no error writing.

response_test.go
```diff
@@ -22,6 +22,7 @@ func TestResponseWriter(t *testing.T) {
 	r.Write([]byte("boop"))
 	b, _ := r.BodyBytes(true)
 	assert.Equal(t, []byte("boop"), b)
+	assert.Equal(t, r.ContentLength, int64(4))
```
suhailpatel: nitpick but can we follow the semantics of `expected`, `actual` (flip the two arguments). This also holds for the other test cases added 🙏

e2e_test.go
```diff
@@ -582,6 +588,8 @@ func TestE2EFullDuplex(t *testing.T) {
 		req.Body = Streamer()
 		defer req.Body.Close()
 		rsp := req.Send().Response()
+		assert.True(t, rsp.ContentLength < 0)
```
suhailpatel: We can assert an exact `-1` contentlength here? There won't be ambiguity

request.go
```diff
@@ -61,9 +61,6 @@ func (r *Request) Encode(v interface{}) {
 		return
 	}
 	r.Header.Set("Content-Type", "application/json")
-	if r.ContentLength < 0 && cw.n < chunkThreshold {
```
suhailpatel: You no longer need the counting writer a few lines above, you can just use `r` directly

response.go
```diff
@@ -96,8 +94,23 @@ func (r *Response) Write(b []byte) (int, error) {
 			rc.Close()
 		}
 		r.Body = buf
-		return buf.Write(b)
+		n, err = buf.Write(b)
+		if err != nil {
+			return 0, err
```
suhailpatel: I don't think this is entirely correct, you can get back an error if `n < len(b)` but that doesn't mean that n is zero. We should return `n, err`

request.go
suhailpatel: With the most recent changes, calling `Encode` won't overwrite `Content-Length` and keep it at zero (note that a zero value is valid as a body length), a `-1` is what triggers chunked encoding. 

We should either set the content length for the json.Marshaler to `-1` or set an explicit length

obeattie: `Request.Write` is called under the hood which entirely takes care of setting ContentLength now, so why do we need to do anything special in the `json.Marshaler` case where we are not replacing `r.Body`?

suhailpatel: Yes, I didn't put the two together. `r` satisfies `io.Writer` and thus implicitly calls `Write` for the encoder.

jackkleeman: As discussed over DM, the json.Marshaler does indeed set content length (and tests verify this)

request.go
```diff
@@ -76,11 +71,14 @@ func (r Request) Decode(v interface{}) error {
...
 // Write writes the passed bytes to the request's body.
-func (r *Request) Write(b []byte) (int, error) {
+func (r *Request) Write(b []byte) (n int, err error) {
 	switch rc := r.Body.(type) {
 	// In the "normal" case, the response body will be a buffer, to which we can write
 	case io.Writer:
-		return rc.Write(b)
+		n, err = rc.Write(b)
+		if err != nil {
+			return 0, err
```
obeattie: This should return `n`, not `0`

response.go
```diff
@@ -75,14 +70,17 @@ func (r *Response) Decode(v interface{}) error {
...
 	if r.Response == nil {
 		r.Response = newHTTPResponse(Request{})
 	}
 	switch rc := r.Body.(type) {
 	// In the "regular" case, the response body will be a bufCloser; we can write
 	case io.Writer:
-		return rc.Write(b)
+		n, err = rc.Write(b)
+		if err != nil {
+			return 0, err
```
obeattie: This should return `n`, not `0`

PR #120, opened 2019-11-01
ineffectual assignment to err

tegk:
It is a bad practice to set variables that will not be used later. Especially in a encoder/decoder that is possibly used frequently.

The variable `err := error(nil)` should be set after the possible return of `r.Error` otherwise it might end up not being used at all.


obeattie:
I don't think there is any logic issue here, but it would save an allocation in the case the first `if` branch matches I guess. So happy to merge, and thanks 👍

PR #121, opened 2020-02-07
[EXPERIMENTAL] Support HTTP2 prior knowledge client

jackkleeman:
Allow the use of an HTTP2 prior knowledge client, as indicated by a context flag.

Prior Knowledge connections are discussed in [section 3.4 of the HTTP/2 RFC](https://tools.ietf.org/html/rfc7540#section-3.4):

```
A client can learn that a particular server supports HTTP/2 by other
means.  For example, [ALT-SVC] describes a mechanism for advertising
this capability.

A client MUST send the connection preface (Section 3.5) and then MAY
immediately send HTTP/2 frames to such a server; servers can identify
these connections by the presence of the connection preface. This
only affects the establishment of HTTP/2 connections over cleartext
TCP; implementations that support HTTP/2 over TLS MUST use protocol
negotiation in TLS [TLS-ALPN].

Likewise, the server MUST send a connection preface (Section 3.5).

Without additional information, prior support for HTTP/2 is not a
strong signal that a given server will support HTTP/2 for future
connections.  For example, it is possible for server configurations
to change, for configurations to differ between instances in
clustered servers, or for network conditions to change.
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,10 +1,14 @@
 package typhon
 
 import (
+	"context"
+	"crypto/tls"
+	"net"
 	"net/http"
 	"time"
 
 	"github.com/monzo/terrors"
+	"golang.org/x/net/http2"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -1,23 +1,39 @@
...
 var (
 	// Client is used to send all requests by default. It can be overridden globally but MUST only be done before use
 	// takes place; access is not synchronised.
 	Client Service = BareClient
 	// RoundTripper is used by default in Typhon clients
-	RoundTripper http.RoundTripper = &http.Transport{
+	// It chooses HTTP1, H2C or HTTP2 based on a context flag (see WithHTTP2)
+	RoundTripper http.RoundTripper = dynamicRoundTripper{}
+	// HTTP1RoundTripper is a HTTP1 client
+	HTTP1RoundTripper http.RoundTripper = &http.Transport{
```
mattheath: Is this definitely a HTTP/1.1 only round tripper? The default `http.Transport` will upgrade to HTTP/2 unless custom TLS config or custom dialers have been set I think?

https://golang.org/src/net/http/transport.go?s=3397:10477#L353
```go
if !t.ForceAttemptHTTP2 && (t.TLSClientConfig != nil || t.Dial != nil || t.DialTLS != nil || t.DialContext != nil) {
```
So in this case `ForceAttemptHTTP2` will default to `false`, but we've not overridden any of the others, so it might still upgrade?

jackkleeman: Yeah, you're right actually, HTTP2 upgrade is already supported. I think we only need an H2C client and can change to `WithH2C`

milesbxf:
Rebased in #146

PR #124, opened 2020-05-05
Set MaxConcurrentStreams to uint32_max

jackkleeman:
We're matching envoy and grpc by making this unlimited. There's no
particular reason to do otherwise, and servers that need to do so can
set up their own http2 server.

### DIFF
--- a/h2c.go
+++ b/h2c.go
@@ -3,6 +3,7 @@ package typhon
 import (
 	"bufio"
 	"context"
+	"math"
 	"net"
 	"net/http"
 	"net/textproto"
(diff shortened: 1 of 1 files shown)


h2c.go
```diff
@@ -116,7 +116,11 @@ func setupH2cHijacker(req Request, rw http.ResponseWriter) (http.ResponseWriter,
 
 	h2c := &h2cInfo{
 		conns: mapset.NewSet(),
-		h2s:   &http2.Server{}}
+		h2s: &http2.Server{
+			// We're copying envoy and grpc by setting this to the max uint32.
+			// The Go default is 250 which is not ideal for long lived streaming requests.
+			MaxConcurrentStreams: 1<<32 - 1,
```
obeattie, line 122: I know this is equivalent so might seem a bit nitpicky, but I think it would read more easily if you set this to `math.MaxUint32`

PR #126, opened 2020-07-23
[BP-513] Add support for wrapping downstream request errors

robinjmurphy:
This PR adds support for wrapping downstream errors in the `Response` `.Decode` method.

Our goal here is to prevent downstream errors from propagating across service boundaries by default. This would mean that downstream `not_found` errors are converted into `internal_service` errors by default (with the underlying `not_found` error preserved as a _cause_ on the newly created error).

The implementation uses a context value to introduce this behaviour in a way that makes it configurable for callers. We'd like to experiment with this behaviour in our own services before recommending it and so we need to be able to enable it on a per-service basis.

We tried first to implement this as a filter, however, when creating a new error within a filter the stack trace doesn't include the entire call stack because the filter is executed in a separate go routine. Adding this code at the point when `Decode` is called ensures that the stack trace includes the entire call stack.

We also considered implementing this behaviour outside of Typhon in the generated client code for each service. However, that requires introducing a config dependency into all of the generated clients and regenerating them.

The approach in this PR feels like the most practical way to enable this feature on a per-service basis, even though it feels like somewhat of an abuse of the context field (to influence control flow in the library).

https://mondough.atlassian.net/browse/BP-513

### DIFF
--- a/response.go
+++ b/response.go
@@ -48,9 +48,21 @@ func (r *Response) Encode(v interface{}) {
 	r.Header.Set("Content-Type", "application/json")
 }
 
+// MaskDownstreamErrors is a context key that can be used to enable
+// wrapping of downstream response errors on a per-request basis.
+//
+// This is implemented as a context key to allow us to migrate individual
+// services from the old behaviour to the new behaviour without adding a
+// dependency on config to Typhon.
+type MaskDownstreamErrors struct{}
+
 // Decode de-serialises the JSON body into the passed object.
 func (r *Response) Decode(v interface{}) error {
 	if r.Error != nil {
+		if s, ok := r.Request.Context.Value(MaskDownstreamErrors{}).(string); ok && s != "" {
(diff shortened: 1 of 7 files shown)


response.go
```diff
@@ -48,9 +48,21 @@ func (r *Response) Encode(v interface{}) {
 	r.Header.Set("Content-Type", "application/json")
 }
 
+// MaskDownstreamErrors is a context key that can be used to enable
+// wrapping of downstream response errors on a per-request basis.
+//
+// This is implemented as a context key to allow us to migrate individual
+// services from the old behaviour to the new behaviour without adding a
+// dependency on config to Typhon.
+type MaskDownstreamErrors struct{}
```
robinjmurphy, line 57: Not sure about `Mask` here but I thought it was the most descriptive. Alternatively `Wrap...`?

Willyham: Mask has a bit of a negative connotation for me so I'd personally go with Wrap, but I don't think it matters too much

robinjmurphy: Yep I'll go with Wrap

PR #127, opened 2020-08-31
Add support for encoding and decoding protobuf messages 📩

suhailpatel:
This PR adds (optional) support for encoding and decoding Protocol Buffer messages within Typhon.

A key tenant of this PR is to maintain the status quo of existing behaviour and to allow for explicit hooks to opt-in to protobuf support
- `typhon.Request` only encodes requests as protobuf if explicitly asked via the `EncodeAsProtobuf` public method or via propagation of a context key 
- `typhon.Response` only encodes responses as protobuf if there is a request associated and the request has an `Accept` header denoting `application/protobuf` support.

Both `typhon.Request` and `typhon.Response` handle decoding support based on the `Content-Type` header. We attempt to decode the response as protobuf if all these points are true
- The `Content-Type` suggests a protobuf messsage (`application/protobuf`, `application/x-protobuf` or `application/octet-stream`)
- The receiver implements `proto.Message`


obeattie:
Thanks so much for breaking ground on this ❤️ As I've said before I'm really keen we make this move to proper Protobuf encoding as I think there are a lot of really important benefits.

I've not reviewed the code in close detail yet, but the bits I have looked at looks good. Thanks especially for putting in such extensive tests, and for figuring out that whole `flavours` thing that could probably be better explained.

The major point of discussion I want to raise is whether calling should need to be concerned with this change at all. At the moment, callers simply `Encode()` and `Decode()` their objects and Typhon kinda does the rest. At least at Monzo, I am fairly sure that we don't have much/any internal RPC code that cares at all whether the request is encoded as JSON or Protobuf. And I think we should aim to transition **all** internal RPC to Protobuf. It would be a shame in my eyes if we ended up with a part-Protobuf-part-JSON situation – we'd be maintaining the complexity of two encoding formats forever. As such I'd challenge whether we should aim for "explicit hooks to opt-in to protobuf support".

So, I wonder if things like `EncodeRequestAsProtobuf` should be made "public". We might still need them for the duration of the transition though, when our own frameworks like Bedrock will still presumably need a way to configure the encoding in use, but perhaps there is some way of doing this that doesn't mean we end up with service code that depends on these kinds of constants. What do you think?

response.go
```diff
@@ -22,15 +24,16 @@ type Response struct {
...
 }
 
-// Encode serialises the passed object as JSON into the body (and sets appropriate headers).
+// Encode serialises the passed object into the body (and sets appropriate headers).
 func (r *Response) Encode(v interface{}) {
 	if r.Response == nil {
 		r.Response = newHTTPResponse(Request{})
 	}
 
+	// If we have a proto message of a
```
tim-monzo, line 33: unsolicited comment: is this a sentence fragment?

suhailpatel: Good spot, removing 👍

obeattie (commented):
Looking good!

request.go
```diff
@@ -61,10 +62,38 @@ func (r *Request) Encode(v interface{}) {
 	r.Header.Set("Content-Type", "application/json")
 }
 
-// Decode de-serialises the JSON body into the passed object.
+// EncodeAsProtobuf serialises the passed object as protobuf into the body
+func (r *Request) EncodeAsProtobuf(m proto.Message) {
```
obeattie, line 66: Is this deliberately exported? I see the counterpart on `Response` is not exported, so curious why it's asymmetrical.

errors.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/golang/protobuf/proto"
```
kjgorman, line 10: this import path was recently deprecated: https://github.com/golang/protobuf/pull/1306/files

errors.go
```diff
@@ -89,14 +90,22 @@ func ErrorFilter(req Request, svc Service) Response {
 		b, _ := rsp.BodyBytes(false)
 		switch rsp.Header.Get("Terror") {
 		case "1":
+			var err error
 			tp := &terrorsproto.Error{}
-			if err := json.Unmarshal(b, tp); err != nil {
+
+			switch rsp.Header.Get("Content-Type") {
+			case "application/octet-stream", "application/x-google-protobuf", "application/protobuf":
```
danielchatfield, line 97: As far as I can see we never set a content type of `application/octet-stream`, is there a particular reason why we want to support it here?

Willyham:
Thanks for looking at this ❤️ . As far as I know `github.com/golang/protobuf/proto.Message` is now deprecated and replaced by `google.golang.org/protobuf/proto.Message` which is a different type entirely. Can we make sure we're using the most up-to-date implementation here?

errors.go
suhailpatel: I wanted to support types commonly sent, [protorpc uses octet-stream](https://github.com/google/protorpc/commit/eb03145a6a7c72ae6cc43867d9635a5b8d8c4545#diff-9409428171f083cd5b68e3037d0f92756b895617effda1a591609e609b4c6f2eR49-R51)

I don't feel too strongly either way. I don't think this makes this path any less correct?

suhailpatel:
> As far as I know github.com/golang/protobuf/proto.Message is now deprecated and replaced by google.golang.org/protobuf/proto.Message which is a different type entirely

Great point, I think we may need to support both (i'll see if that's something we can do here). Internally we use the former and upgrading is not as trivial as it sounds (and it'd be nice to decouple the two things if we can avoid it).

request.go
```diff
@@ -62,10 +63,38 @@ func (r *Request) Encode(v interface{}) {
...
+		r.err = terrors.Wrap(err, nil)
+		return
+	}
+
+	n, err := r.Write(out)
+	if err != nil {
+		r.err = terrors.Wrap(err, nil)
+		return
+	}
+	r.Header.Set("Content-Type", "application/x-protobuf")
```
theninj4, line 79: This smells like it could be wrong (or incompatible)?

When we encode, we set a Content-Type of `application/x-protobuf`.
When we decode we look for a range Content-Types, none of which include the one above.

suhailpatel: We [check for application/x-protobuf here](https://github.com/monzo/typhon/pull/127/files#diff-3c2feffcf57ffb6c596dc5608186c518252092b7017be5189d3dc12aa3c45cecR95-R100)

errors.go
danielchatfield: Yeh, it's probably fine. I think what made me question it is thinking that it's really easy to add later but can be quite difficult to remove and unlike the others this content type isn't really linked to proto – there are lots of things that would use this content type.

request.go
```diff
@@ -62,11 +64,50 @@ func (r *Request) Encode(v interface{}) {
...
+			return terrors.InternalService("invalid_type", "could not decode proto message", nil)
+		}
+		err = proto.Unmarshal(b, m)
+
+	// Proper JSON handling requires the protojson package in Go. application/jsonpb is a suggestion by grpc-gateway:
+	// https://github.com/grpc-ecosystem/grpc-gateway/blob/f4371f7/runtime/marshaler_registry.go#L89-L90
+	// This is a backward compatibility break for those using google.golang.org/protobuf/proto.Message incorrectly.
+	default:
+		if m, ok := v.(proto.Message); ok {
+			err = protojson.Unmarshal(b, m)
```
danielchatfield, line 107: I'd like to understand this change a bit better. Typhon doesn't encode proto messages as proto compliant JSON messages, it calls `json.Marshal`. When we decode do we not also need to decode using `json.Unmarshal` to handle this correctly?


I wonder if we could add a test that uses a language feature for which the encoding differs between `json.Marshal` and the proto spec and assert that we can encode and decode successfully. For example I believe enums are serialised differently.

intUnderflow: On an aside to this: Perhaps we should have some tests check decoding of old marshalled JSON returns expected values, so we don't accidentally break compatibility in the future

suhailpatel:
To compile the prototest proto
```
$ cd $GOPATH/src/github.com/monzo/typhon
$ protoc --proto_path=./prototest prototest/*.proto --go_out=prototest/
```

request.go
intUnderflow: I think you're right, protojson != json so we'll need to handle it separately, I'll add a hint to the content-type to indicate whether we're using protojson or not

We can also add a test to make sure old JSON works properly :)

intUnderflow (approved):

intUnderflow:
Integrated #141 and rebased

PR #133, opened 2021-04-09
Add ResponseWithCode method to Typhon request

campesel:
While reading [the proposal for API standards at Monzo](https://www.notion.so/monzo/API-Standards-4f1268b9bb3741749939ea7f16466fd5), I noticed we don't have a method to construct a Typhon response to a request _and_ set its HTTP status code at the same time.

This means that to return specific HTTP status codes from our APIs, the application code needs to set the status code on the response returned by the `Response` method.

For example, the new API standards suggest returning `201 Created` in response to a `PUT` request that creates a new entity. To do this with Typhon, we would write something along these lines:

```
rsp := req.Response(&MyResponseType{})
rsp.StatusCode = http.StatusCreated
return rsp
```

This PR adds a method `ResponseWithCode` to `Request`, so that we could do this instead:

```
return req.Response(&MyResponseType{}, http.StatusCreated)
```

If you feel this could be useful, I'd be happy add comments and tests. Otherwise, if there are reasons not to do this, I'll just close the PR!

### DIFF
--- a/request.go
+++ b/request.go
@@ -147,9 +147,19 @@ func (r Request) SendVia(svc Service) *ResponseFuture {
 	return SendVia(r, svc)
 }
 
-// Response construct a new Response to the request, and if non-nil, encodes the given body into it.
+// Response constructs a new Response to the request, and if non-nil, encodes the given body into it.
 func (r Request) Response(body interface{}) Response {
 	rsp := NewResponse(r)
+	if body != nil {
+		rsp.Encode(body)
+	}
+	return rsp
+}
+
+// ResponseWithCode constructs a new Response with the given status code to the request, and if non-nil, encodes the
+// given body into it.
+func (r Request) ResponseWithCode(body interface{}, statusCode int) Response {
+	rsp := NewResponseWithCode(r, statusCode)
 	if body != nil {
 		rsp.Encode(body)
 	}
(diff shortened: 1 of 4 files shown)


Willyham:
Seems like a useful change to me 👍

response.go
```diff
@@ -189,10 +189,18 @@ func newHTTPResponse(req Request) *http.Response {
 		Body:          &bufCloser{}}
 }
 
-// NewResponse constructs a Response
+// NewResponse constructs a Response with status code 200.
 func NewResponse(req Request) Response {
 	return Response{
```
Willyham, line 194: Should this just `return NewResponseWithCode(...)`?

request.go
```diff
@@ -147,7 +147,7 @@ func (r Request) SendVia(svc Service) *ResponseFuture {
 	return SendVia(r, svc)
 }
 
-// Response construct a new Response to the request, and if non-nil, encodes the given body into it.
+// Response constructs a new Response to the request, and if non-nil, encodes the given body into it.
 func (r Request) Response(body interface{}) Response {
 	rsp := NewResponse(r)
```
Willyham, line 152: This could also call `r.ResponseWithCode(body, http.StatusOK)`

campesel: I think I'd prefer to have `Response` call `NewResponse` (which defaults to `http.StatusOK`) and `ResponseWithCode` call `NewResponseWithCode` with the given status code.

If you feel strongly about this, I'm happy to change it!

Willyham (approved):

PR #134, opened 2021-04-11
Enable aborting streamer with error

rdingwall:
If a server is using `typhon.Streamer()` and encounters an error that prevents it from writing the next chunk of data to the response (e.g. failed to read the next record from the database), there is currently no way for the server to signal to the client that streaming terminated unsuccessfully. (Besides simply stopping writing and not calling `streamer.Close()` so the client times out (and leaking a goroutine on the server), or encoding some sort of application-level error into the response body). 

For example:

```go
func streamingService(req typhon.Request) typhon.Response {
    body := typhon.Streamer()
    go func() {
        for {
            data, err := readNextData(...)
            if err != nil {
                // what should I do here?
            }
        }
        body.Close()
    }()
    return req.Response(body)
}
```

Not having an obvious blessed path for aborting a streaming response also creates a risk of engineers falling into this pitfall, which would result in the client reading a truncated response:

```go
func streamingService(req typhon.Request) typhon.Response {
    body := typhon.Streamer()
    go func() {
        defer body.Close()
        for {
            data, err := readNextData(...)
            if err != nil {
                return
            }
        }
    }()
    return req.Response(body)
}
```

To fix this, this PR adds a new `streamer.CloseWithError(error)` method, which aborts the HTTP response:

```go
func streamingService(req typhon.Request) typhon.Response {
    body := typhon.Streamer()
    go func() {
        for {
            data, err := readNextData(...)
            if err != nil {
                body.CloseWithError(err)
                return
            }
        }
        body.Close()
    }()
    return req.Response(body)
}
```

Internally, typhon aborts the HTTP response by calling `panic(http.ErrAbortHandler)`, which is the official way to abort an HTTP response within `ServeHTTP` (for HTTP 1.1 and 2). See https://github.com/golang/go/issues/17790.

One thing I'm not sure about is, is it safe to assume the panic would always be caught by `http.conn.serve()` [here](https://github.com/golang/go/blob/master/src/net/http/server.go#L1818)? Is there any risk the panic could be intercepted by some intermediate function (e.g. a filter)?

To expose the `CloseWithError()` method, I had to change `typhon.Streamer()` from returning an `io.ReadWriteCloser` to a `typhon.StreamWriter`. Changing the return type is a breaking change (e.g. if a caller passes the streamer to a function), but I couldn't see any other non-nasty way of doing this.

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


e2e_test.go
```diff
@@ -561,6 +562,64 @@ func TestE2EStreamingCancellation(t *testing.T) {
...
+		ctx := context.Background()
+		req := NewRequest(ctx, "GET", flav.URL(s), nil)
+		rsp := req.Send().Response()
+		<-done
+		body, err := ioutil.ReadAll(rsp.Body)
+		assert.Equal(t, "derp\n", string(body))
+		assert.EqualError(t, err, io.ErrUnexpectedEOF.Error())
+	})
+
+	someFlavours(t, []string{"http2.0-h2", "http2.0-h2c"}, func(t *testing.T, flav e2eFlavour) {
```
rdingwall, line 594: There are different flavors for HTTP 1.1 and 2 because they return different errors.

obeattie (approved):
Sorry for taking my sweet time in reviewing this. In summary: this is great and I'm very in favour. Thank for doing it! 🙌 I actually didn't know that aborting a streaming response like this was possible with Go's HTTP package before. So, TIL.

>is it safe to assume the panic would always be caught by `http.conn.serve()` [here](https://github.com/golang/go/blob/master/src/net/http/server.go#L1818)? Is there any risk the panic could be intercepted by some intermediate function (e.g. a filter)?

I _think_ that's a safe assumption, yeah. This panic gets raised inside `HttpHandler` itself, which means it should be outside the filter stack since that stack gets invoked by that function.

PR #141, opened 2021-11-22
Encoding / Decoding changes in protobuf branch

intUnderflow:
This PR does a few things:
1. In response to Daniel's comment on the protobuf PR [here](https://github.com/monzo/typhon/pull/127#discussion_r750562509), this makes Typhon use only regular JSON or protobufs
2. ~~Adds new EncodeAsProtoJSON to encode a message as protojson~~
3. Maps Encode to point at EncodeAsProtobuf
4. ~~Blocks requests with a non-standard content-type from Typhon~~

### DIFF
--- a/request.go
+++ b/request.go
@@ -41,10 +41,8 @@ func (r *Request) unwrappedContext() context.Context {
 	}
 }
 
-// Encode maps to to EncodeAsJSON
-// TODO: Remove in the next major release and require encoding to explicitly go through either EncodeAsJSON, EncodeAsProtoJSON or EncodeAsProtobuf
 func (r *Request) Encode(v interface{}) {
-	r.EncodeAsJSON(v)
+	r.EncodeAsProtobuf(v)
 }
 
 // EncodeAsJSON serialises the passed object as JSON into the body (and sets appropriate headers).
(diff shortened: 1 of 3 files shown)


intUnderflow:
I'm happy to take out 4 and leave for a major release as its a massive breaking change BTW

danielchatfield:
>I'm happy to take out 4 and leave for a major release as its a massive breaking change BTW

Yeh, that was going to be my comment – I think that's a larger change that should be separated.

request.go
```diff
@@ -64,7 +70,24 @@ func (r *Request) Encode(v interface{}) {
 	r.Header.Set("Content-Type", "application/json")
 }
 
-// EncodeAsProtobuf serialises the passed object as protobuf into the body
+// EncodeAsProtoJSON serialises the passed object as ProtoJSON into the body (and sets appropriate headers).
+func (r *Request) EncodeAsProtoJSON(m proto.Message) {
```
suhailpatel, line 74: I wonder if we should expose this as a public function at all? I can't think of a use-case where we'd want to explicitly send out protojson delibrately. We need to continue to support in the Decode but I don't believe we should explicitly add a new method to Encode

If we do intend to keep this in the interface, can we add some explicit tests please so we don't have a further regression in future 🙏

intUnderflow: I've removed this endpoint and just left legacy JSON / protobuf

suhailpatel (approved):

PR #143, opened 2022-01-26
Requests should allow encoding and decoding as legacy protobuf

suhailpatel:
In some cases, we use the old protobuf and we need to support encoding and decoding of requests as that type


request.go
```diff
@@ -97,19 +98,18 @@ func (r Request) Decode(v interface{}) error {
...
+		case legacyproto.Message:
+			err = legacyproto.Unmarshal(b, m)
+		default:
 			return terrors.InternalService("invalid_type", "could not decode proto message", nil)
 		}
-		err = proto.Unmarshal(b, m)
 	// As older versions of typhon used json, we don't use protojson here as they are mutually exclusive standards with
 	// major differences in how they handle some types (such as Enums)
 	default:
-		m, ok := v.(proto.Message)
```
suhailpatel, line 108: This was a regression, if a value comes in and we can't decide if it's a protobuf message, we should attempt to json decode it without any inference

intUnderflow: 👍🏻 let's release this as 1.0.1 in that case

intUnderflow (approved):
LGTM 😄 thanks for getting this done

PR #148, opened 2022-04-27
Add basic circleci configuration for typhon

cstorey-monzo:
I've based this on both the existing [`.travis.yml`](https://github.com/monzo/typhon/blob/master/.travis.yml) and [wearedev's  CircleCI setup](https://github.com/monzo/wearedev/blob/90cdffb637faa3325406810a857bb0821c147ed0/.circleci/config.yml). 

Granted, that's the easy part, and it's very much free as in puppy. I'm not currently sure about the `$PATH` is setup, so I think that the first build might fail because it can't find `golint`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,23 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/golang:1.17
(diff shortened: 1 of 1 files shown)


cstorey-monzo:
I'm told by @suhailpatel that we prefer Github actions for this case, so I'll close this for now.

PR #149, opened 2022-04-27
Add "Hello world" configuration.

cstorey-monzo:
This will mean we can turn on CircleCI for the typhon repository without build failures on main, while we get https://github.com/monzo/typhon/pull/148 working.

This is one of the sample configurations given by CircleCI itself.


cstorey-monzo:
I'm told by @suhailpatel that we prefer Github actions for this case, so I'll close this for now.

PR #151, opened 2022-05-09
Log http2: stream closed as debug-level

milesbxf:
We had some internal reports of services logging warnings like "Couldn't send response body", with the underlying error being "http2: stream closed".

These errors occur as a result of the client closing the stream as we're still writing to it, which is a frequent and expected part of HTTP/2 operation, so we can safely log these at debug level.

### DIFF
--- a/http.go
+++ b/http.go
@@ -66,7 +66,8 @@ func copyErrSeverity(err error) slog.Severity {
 
 	switch {
 	case strings.HasSuffix(err.Error(), "read on closed response body"),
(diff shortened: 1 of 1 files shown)


domgoodwin (approved):
LGTM 🚀 
Are there any other potential messages we need to add to this list for http2?

milesbxf:
> LGTM 🚀 Are there any other potential messages we need to add to this list for http2?

I don't think so off the top of my head - we can always come back and add them later

PR #152, opened 2022-05-12
Add support for server timeouts

milesbxf:
This allows configuring the standard Go http.Server timeouts on Typhon servers, and adds a new ServerOption interface that we can use to extend Typhon config in the future.

Sadly I couldn't get this working for h2c - this is due to a bug in Go where the original *http.Server configuration doesn't get propagated to h2c connections. See https://github.com/golang/go/issues/52868

### DIFF
--- a/e2e_http1_test.go
+++ b/e2e_http1_test.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"testing"
+
 	"github.com/monzo/terrors"
 	"github.com/stretchr/testify/assert"
-	"testing"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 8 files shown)


e2e_http1_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/monzo/terrors"
```
suhailpatel, line 7: nit: can we make sure the imports are appropriately grouped

suhailpatel (approved):
LGTM but please address nit comment

e2e_http1_test.go
milesbxf: Urgh I think my Goland imports setup for this repo is buggered, let me take a look

suhailpatel (approved):

PR #153, opened 2022-06-06
Don't write body for certain status codes

jace-ys:
[Certain HTTP status codes don't allow writing a body](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/net/http/transfer.go;l=456-466;drc=d8762b2f4532cc2e5ec539670b88bbc469a13938), so we should check for these same set of
status codes before attempting to write the body (in [`copyChunked`](https://github.com/monzo/typhon/blob/master/http.go#L122) and [`io.CopyBuffer`](https://github.com/monzo/typhon/blob/master/http.go#L131)) to prevent
unnecessarily logging an error when the call to `Write` returns `http.ErrBodyNotAllowed`.

### DIFF
--- a/http.go
+++ b/http.go
@@ -87,6 +87,20 @@ func copyErrSeverity(err error) slog.Severity {
 	}
 }
 
+// bodyAllowedForStatus reports whether a given response status code
+// permits a body. This is taken directly from the net/http package.
+func bodyAllowedForStatus(status int) bool {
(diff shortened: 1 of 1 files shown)


milesbxf (approved):
Thanks for this 🙏  agree it doesn't make sense to panic if we hit this condition

PR #158, opened 2022-10-31
error-filter: Propogate body read errors

EddShaw:
If we fail to read the entire body, we should propogate that error.

This will allow us to stop emitting this I/O error to Sentry.

Don't attempt to recover the partially read bytes - they might contain sensitive data, and putting them in a terror risks them being propogated to an unintended target.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -106,7 +106,7 @@ func ErrorFilter(req Request, svc Service) Response {
 		b, err := rsp.BodyBytes(false)
 		if err != nil {
 			// Don't attempt to parse a partially read error response. Return the underlying read error when this occurs.
-			rsp.Error = terrors.Wrap(err, nil)
+			rsp.Error = terrors.NewInternalWithCause(err, "reading error response body", nil, "body_read_error")
 		} else {
 			switch rsp.Header.Get("Terror") {
 			case "1":
--- a/errors.go
+++ b/errors.go
@@ -1,11 +1,13 @@
 package typhon
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+	"unicode/utf8"
 
 	legacyproto "github.com/golang/protobuf/proto"
 	"github.com/monzo/slog"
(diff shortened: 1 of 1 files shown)


errors.go
```diff
@@ -103,27 +103,32 @@ func ErrorFilter(req Request, svc Service) Response {
...
 		// There is an error in the underlying response; unmarshal
-		b, _ := rsp.BodyBytes(false)
-		switch rsp.Header.Get("Terror") {
-		case "1":
-			var err error
-			tp := &terrorsproto.Error{}
+		b, err := rsp.BodyBytes(false)
+		if err != nil {
+			// Don't attempt to parse a partially read error response. Return the underlying read error when this occurs.
+			rsp.Error = terrors.Wrap(err, nil)
```
WillSewell, line 109: A couple of suggestions here:

- avoid terrors.Wrap because it's deprecated
- include any body we did read in the error
- include a specific terrors sub code to help distinguish this error in e.g. Sentry filter list

i.e.

```
errParams := map[string]string{
  "partially_read_body": b,
}
terrors.NewInternalWithCause(err, "failed to read response body", errParams, "body_read_failed")
```

WillSewell (changes requested):

errors.go
EddShaw: > avoid terrors.Wrap because it's deprecated

Not in the version of terrors used by typhon. 🤣 NewInternalWithCause seems sensible though. 👌 


> include any body we did read in the error

I deliberately _didn't_ include the body - with the reasoning in the PR description:
> Don't attempt to recover the partially read bytes - they might contain sensitive data, and putting them in a terror risks them being propogated to an unintended target.

Very happy to be challenged on this (especially as this is the existing state of the world). Thoughts?


> include a specific terrors sub code to help distinguish this error in e.g. Sentry filter list

Good idea.

WillSewell: > Don't attempt to recover the partially read bytes - they might contain sensitive data, and putting them in a terror risks them being propogated to an unintended target.

I might be missing something because I don't really get the concern. If the error body _is_ successfully read, then we include the entire body in the error. In which case, why not also do this for a partially read body?

EddShaw: > I might be missing something because I don't really get the concern. If the error body is successfully read, then we include the entire body in the error. In which case, why not also do this for a partially read body?

To be honest, there are situations where this makes me uncomfortable too.

We already encourage against putting sensitive data in terrors? So for Monzo RPCs, this is moot.

But I've seen we also use Typhon for other Go-originated HTTP in `wearedev`, including calling 3rd party services that don't have an official Go client. External APIs are where this worries me most. On external APIs, we can reasonably expect:
* Non-200 status codes are used in expected responses
    * e.g. Stripe does this a lot. It'll return 400 status response for failing to take a payment on a PaymentIntent - with the complete Payment Intent object in the body 
* PII and secrets can be returned in these responses
    * Stripe's PaymentIntent client_secret field should not be logged or be visible to anyone except the customer.

(For Stripe, we're already using the official Go client - it's just the HTTP API that I've been working with recently 😅)

I think you're right - that the non-200 response body wasn't entirely parsed isn't actually significant. Typhon already makes it hard to get to the response body in this case, so it's unlikely that we're parsing this error content today. Typhon + `ErrorFilter` shouldn't be our default choice of HTTP client for 3rd party APIs. Conflating I/O errors with non-200 responses is very convenient for Monzo RPCs, but is a foot-gun with HTTP in general.

EddShaw: Added the `partially_read_body` error param (including handling when the param isn't valid utf8). @WillSewell

WillSewell (approved):

errors.go
WillSewell: Yeah I can see your concern here - though I think that's a wider issue than the change you are making in this PR, so we can defer for now.

WillSewell:
(we shouldn't get the circleci failures - I think I've disabled this for future PRs)

PR #159, opened 2023-01-19
Deprecate BodyBytes functions

WillSewell:
We have been bitten by subtle bugs that are caused by this function being called on a value (and not a pointer to a value), combined with `consume == false`.

Because of this we think the convenience of these functions are not worth the risk.


WillSewell:
Decided to just warn instead: https://github.com/monzo/typhon/pull/160

PR #161, opened 2023-02-23
Use DialTLSContext instead of DialTLS when negotiating HTTP2 session

joechrisellis:
... so that we respect context deadlines when we're dialling.

The old DialTLS is marked as deprecated now anyways.


joechrisellis:
... we're blocked on #162 for now, will rebase once that's landed. 🙂

PR #166, opened 2023-05-05
Add max connection age server option

danielchatfield:
This pull request adds a new `typhon.ServerOption` that will enforce a server-side maximum connection age. This is useful in environments where long-lived connections can result in unbalanced load.

It's a little frustrating that I haven't been able to write a really robust test for this, I've had to settle for validating that the code is executed and the connection start time is being propagated. In practice this is a pretty decent test since the only thing it isn't covering is our ability to correctly determine if that connection start is > maxAge and to set a header which are both simple.

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -14,10 +14,11 @@ import (
 	"time"
 
 	"github.com/fortytw2/leaktest"
-	"github.com/monzo/terrors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/http2"
+
+	"github.com/monzo/terrors"
 
 	"github.com/monzo/typhon/prototest"
 )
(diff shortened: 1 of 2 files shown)


e2e_test.go
```diff
@@ -840,3 +841,39 @@ func TestE2EServerTimeouts(t *testing.T) {
...
+func TestE2EMaxConnectionAge_ConnectionAgeSet(t *testing.T) {
+	addConnectionStartTimeHeader = true
+
+	someFlavours(t, []string{
+		"http1.1",
+		"http1.1-tls",
+		"http2.0-h2",
+
+		// The Go h2c implementation doesn't currently support max
+		// connection age.
```
danielchatfield, line 859: Same as the other server options – not supported by h2c (which we don't use anywhere).

adlawson (approved):

PR #169, opened 2023-12-01
Add RequestPathPattern method to Typhon Request

RohanPadmanabhan:
## 📚 Background
We have found several use cases such as metrics, logging, and authorization where knowing the router path (without any dynamic parameters) is useful to avoid unbounded cardinality or make searching for a particular value easier. At the moment, it is doable to get this but it's not easy. 

## 🆕 What does this PR do?
1. Adds a `RouterEndpointPattern` method to get the router endpoint pattern
2. Adds a `RequestMethod` method to get the request method. This is needed to fulfil an interface between the two APIs. I couldn't name it just `Method` as this clashes with the string inside the contained http.Request object

## 🛡️ Testing
Added new unittests.

### DIFF
--- a/request.go
+++ b/request.go
@@ -215,7 +215,7 @@ func (r Request) ResponseWithCode(body interface{}, statusCode int) Response {
 
 // RequestPathPattern finds the router entry pattern that matches the request
 func (r Request) RequestPathPattern() string {
-	return routerEntryPathPatternForRequest(r)
+	return routerPathPatternForRequest(r)
 }
 
 // RequestMethod returns the HTTP method of the request
--- a/router.go
+++ b/router.go
@@ -46,7 +46,7 @@ func RouterForRequest(r Request) *Router {
 	return nil
 }
 
-func routerEntryPathPatternForRequest(r Request) string {
+func routerPathPatternForRequest(r Request) string {
 	if v := r.Context.Value(routerRequestPatternContextKey); v != nil {
 		return v.(string)
 	}


request.go
```diff
@@ -213,6 +213,22 @@ func (r Request) ResponseWithCode(body interface{}, statusCode int) Response {
...
+func (r Request) RouterEndpointPattern() string {
+	if router := RouterForRequest(r); router != nil {
+		if pathPattern := router.Pattern(r); pathPattern != "" {
+			return pathPattern
+		}
+	}
+	return ""
+}
+
+// RequestMethod returns the HTTP method of the request
```
RohanPadmanabhan, line 227: I'm not particularly keen on the name of this so any suggestions would be appreciated 🙇

request.go
```diff
@@ -213,6 +213,22 @@ func (r Request) ResponseWithCode(body interface{}, statusCode int) Response {
 	return rsp
 }
 
+// RouterEndpointPattern finds the router pattern that matches the request. This is only callable while the request
+// is being served.
+func (r Request) RouterEndpointPattern() string {
```
priyeshpatel, line 218: `RequestMethod` feels intuitive and simple. 👍 

For this method, [`RouterEntryPattern`](https://github.com/monzo/typhon/blob/ea17ae8b9d577dfbeeedf388993b22ba52867da5/router.go#L24) is probably the most correct, but `RequestPathPattern` or even `RequestPattern` would be simplest?

request.go
```diff
@@ -213,6 +213,22 @@ func (r Request) ResponseWithCode(body interface{}, statusCode int) Response {
...
+	if router := RouterForRequest(r); router != nil {
+		if pathPattern := router.Pattern(r); pathPattern != "" {
+			return pathPattern
+		}
+	}
```
priyeshpatel, line 223: An alternative implementation here would be to store the pattern used to serve the request in the context, [as we do with the router currently](https://github.com/monzo/typhon/blob/ea17ae8b9d577dfbeeedf388993b22ba52867da5/router.go#L119-L126). Then we don't need to look up the pattern separately (which could theoretically have changed if a new route is registered in the intervening time).

router.go
```diff
@@ -13,10 +13,12 @@ import (
 // directly means we'd get a collision with any other package that does the same.
 // https://play.golang.org/p/MxhRiL37R-9
 type routerContextKeyType struct{}
+type routerPathContextKeyType struct{}
 
 var (
-	routerContextKey   = routerContextKeyType{}
-	routerComponentsRe = regexp.MustCompile(`(?:^|/)(\*\w*|:\w+)`)
+	routerContextKey     = routerContextKeyType{}
+	routerPathContextKey = routerPathContextKeyType{}
```
priyeshpatel, line 20:
```suggestion
	routerRequestPatternContextKey = routerRequestPatternContextKeyType{}
```

Optional: This might be clearer to distinguish between path and pattern?

priyeshpatel (approved):

router.go
```diff
@@ -44,6 +46,13 @@ func RouterForRequest(r Request) *Router {
 	return nil
 }
 
+func routerEntryPathPatternForRequest(r Request) string {
```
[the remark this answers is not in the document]
RohanPadmanabhan: Renamed in https://github.com/monzo/typhon/pull/169/commits/fc0f84099391b5239090ae39d95cff44a2930683

priyeshpatel (approved):

